PR #2, opened 2019-10-13
Migrate to .net core 2.2

jonasanker:
Upgraded project and dependencies to .net core 2.2 . The upgrade is preparation for that we at some point can move to .net core 3.0


jonasanker:
@TTA777 , @Rolfenz  I suggest that you only look at the first commit (about the upgrade) https://github.com/AnalogIO/analog-core/pull/2/commits/a9ff3e6d6d23ec0abb8bfb35b8eebcf7b2f1e363 . 

This is the most interesting one. The others was just code cleanup which performed a lot but not interesting changes :)

jonasanker:
Please also review the last commit on NSwag https://github.com/AnalogIO/analog-core/pull/2/commits/0c3af13afcade18ad237278dc5a74dd0049226be

PR #3, opened 2019-10-15
Feature/klar/app config

Rolfenz:
Added AppConfig interface/controller/service/DTO

### DIFF
--- a/coffeecard/Controllers/AppConfigController.cs
+++ b/coffeecard/Controllers/AppConfigController.cs
@@ -15,7 +15,7 @@ namespace coffeecard.Controllers
     [ApiController]
     public class AppConfigController : ControllerBase
     {
-        private IAppConfigService _appConfigService;
+        private readonly IAppConfigService _appConfigService;
 
         public AppConfigController(IAppConfigService appConfigService)
         {
@@ -26,7 +26,6 @@ namespace coffeecard.Controllers
         public ActionResult<IEnumerable<AppConfigDTO>> Get()
         {
             var appConfig = _appConfigService.RetreiveConfiguration();
-            if(appConfig == null) return NotFound();
             return Ok(appConfig);
         }
     }
--- a/coffeecard/Controllers/AppConfigController.cs
+++ b/coffeecard/Controllers/AppConfigController.cs
@@ -1,0 +1,33 @@
+﻿using coffeecard.Models.DataTransferObjects.AppConfig;
+using coffeecard.Services;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace coffeecard.Controllers
+{
+    [ApiVersion("1")]
+    [Route("api/v{version:apiVersion}/[controller]")]
+    [Authorize]
+    [ApiController]
+    public class AppConfigController : ControllerBase
+    {
+        private IAppConfigService _appConfigService;
+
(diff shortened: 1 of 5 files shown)


coffeecard/Controllers/AppConfigController.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace coffeecard.Controllers
+{
+    [ApiVersion("1")]
+    [Route("api/v{version:apiVersion}/[controller]")]
+    [Authorize]
+    [ApiController]
+    public class AppConfigController : ControllerBase
+    {
+        private IAppConfigService _appConfigService;
```
jonasanker: pretty comment; make it `readonly` (Java's final)

coffeecard/Controllers/AppConfigController.cs
```diff
@@ -0,0 +1,33 @@
...
+        public AppConfigController(IAppConfigService appConfigService)
+        {
+            _appConfigService = appConfigService;
+        }
+
+        [HttpGet]
+        public ActionResult<IEnumerable<AppConfigDTO>> Get()
+        {
+            var appConfig = _appConfigService.RetreiveConfiguration();
+            if(appConfig == null) return NotFound();
```
jonasanker: Based on your checks in the `AppConfigService` this null check is not needed.

coffeecard/Services/AppConfigService.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class AppConfigService : IAppConfigService
+    {
+        private readonly IConfiguration _configuration;
+
+        public AppConfigService(IConfiguration configuration)
+        {
+            _configuration = configuration;          
+        }
+        //TODO throw exception if EnvironmentType and MerchantID is null or empty-string.
```
jonasanker: clean up comment: remove this resolved todo

coffeecard/Services/AppConfigService.cs
```diff
@@ -0,0 +1,33 @@
...
+        public AppConfigDTO RetreiveConfiguration()
+        {
+            var _environmentType = _configuration["EnvironmentType"];
+            var _merchantId = _configuration["MobilePayMerchantId"];
+            if(string.IsNullOrEmpty(_environmentType)) throw new KeyNotFoundException();
```
jonasanker: I would probably use the `ArgumentNullException` here or make a custom one like `PropertyNotFoundException` since the key is known but it is the value which is missing

coffeecard/Services/AppConfigService.cs
```diff
@@ -0,0 +1,33 @@
...
+            if(string.IsNullOrEmpty(_merchantId)) throw new KeyNotFoundException();
```
jonasanker: Pretty comment; make it a oneliner
```csharp
if(string.IsNullOrEmpty(_environmentType) || string.IsNullOrEmpty(_merchantId)) throw new ArgumentNullException();
```

coffeecard/Services/AppConfigService.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public AppConfigService(IConfiguration configuration)
+        {
+            _configuration = configuration;          
+        }
+        //TODO throw exception if EnvironmentType and MerchantID is null or empty-string.
+        public AppConfigDTO RetreiveConfiguration()
+        {
+            var _environmentType = _configuration["EnvironmentType"];
+            var _merchantId = _configuration["MobilePayMerchantId"];
```
jonasanker: So apparently we already store the value under the key `MPMerchantID`. Use that instead :)

PR #4, opened 2020-01-20
Support new MobilePay flow

jonasanker:


coffeecard/CoffeeCard/Helpers/MobilePay/ErrorMessage/InternalServerErrorMessage.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class InternalServerErrorMessage : IMobilePayErrorMessage
+    {
+        public string CorrelationId { get; set; }
+        public string Errortype { get; set; }
+        public string Message { get; set; }
+
+        public string GetErrorMessage()
+        {
+            return $"CorrelationId = {CorrelationId}, Errortype = {Errortype}, Message = {Message}";
```
TTA777: Why is this method and ToString exactly the same?

coffeecard/CoffeeCard/Helpers/MobilePay/MobilePayException.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Net;
+using CoffeeCard.Helpers.MobilePay.ErrorMessage;
+
+namespace CoffeeCard.Helpers.MobilePay
+{
+    [Serializable]
+    public class MobilePayException : Exception
+    {
+        private readonly IMobilePayErrorMessage _errorMessage;
```
TTA777: It is not possible to use the existing Message field in exceptions in some manner, to avoid confusion when working with the exception?

coffeecard/CoffeeCard/Helpers/MobilePay/ResponseMessage/IMobilePayAPIResponse.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace CoffeeCard.Helpers.MobilePay.ResponseMessage
+{
+    public interface IMobilePayAPIResponse
+    {
```
TTA777: Any reason TransactionId is not in the interface? It seems to be a field that all response messages shares

coffeecard/CoffeeCard/Services/AccountService.cs
```diff
@@ -48,57 +47,41 @@ public string Login(string username, string password, string version)
...
-                    {
-                        new Claim(ClaimTypes.Email, username), new Claim(ClaimTypes.Name, user.Name),
-                        new Claim("UserId", user.Id.ToString())
-                    };
+                    var claims = new Claim[] { new Claim(ClaimTypes.Email, username), new Claim(ClaimTypes.Name, user.Name), new Claim("UserId", user.Id.ToString()) };
                     var token = _tokenService.GenerateToken(claims);
 
-                    // check for incomplete purchases
-                    _purchaseService.CheckIncompletePurchases(user);
+                    // check for incomplete purchases //TODO Fix this and reimplement
```
TTA777: Maybe we want to consider reimplementing this, or removing it altogether if we consider it redundant

coffeecard/CoffeeCard/Helpers/MobilePay/ErrorMessage/InternalServerErrorMessage.cs
jonasanker: I am not sure what was useful to get, so just duplicated for now. Any suggestions?

coffeecard/CoffeeCard/Helpers/MobilePay/MobilePayException.cs
jonasanker: The super class takes a string in it's constructor where we already store the error message. In the `GetErrorMessage()` we return the same error message which we already set in the super class, so I guess youre right there is no need to have it as a field :)

coffeecard/CoffeeCard/Helpers/MobilePay/ResponseMessage/IMobilePayAPIResponse.cs
jonasanker: We cant have fields in interfaces, so we could consider having some inheritance with a super classes or abstract class but personal opinion is when we have to model the objects the MobilePay API is responding with better just keep them exactly as they are.

But I get that the need of an  empty interface is a little weird since it has methods which mus implemented. Motivation for the interface is only to constraint that response messages must be of this type

coffeecard/CoffeeCard/Services/AccountService.cs
jonasanker: I'll think about it - what is the case or scenario where clipcards have not been delivered seen from an API perspective? Wouldnt the API not always deliver or if not able to implement a cancel reservation function?

coffeecard/CoffeeCard/Helpers/MobilePay/ErrorMessage/InternalServerErrorMessage.cs
TTA777: I think the output is fine, maybe just call the toString method in getError message to avoid duplication?

coffeecard/CoffeeCard/Services/AccountService.cs
TTA777: The only thing I can think of is if some unhandled error occurred while issuing the product, though I am not sure if it could/ should recover by itself in such an instance

jonasanker:
Yea. Same thought that could be some sort of database error but if that's
the case, we should retry and it not successful cancel the payment
reservation.

coffeecard/CoffeeCard/Helpers/MobilePay/ResponseMessage/IMobilePayAPIResponse.cs
TTA777: Since you are using getters and setters for it you should be able to add those if we wanted to. I do get your reasoning though, so maybe it's better left as is. Up to you.

PR #7, opened 2020-02-24
Support BaristaOnly products

jonasanker:
- User entity expanded with isBarista boolean
- ProductsService filters on whether the user can see a baristaOnly product
- EF Migration has not been done yet. Awaits deployment to beta environment as well

### DIFF
--- a/coffeecard/CoffeeCard.Tests.Unit/Services/ProductServiceTest.cs
+++ b/coffeecard/CoffeeCard.Tests.Unit/Services/ProductServiceTest.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using CoffeeCard.Configuration;
 using CoffeeCard.Models;
 using CoffeeCard.Services;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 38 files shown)


coffeecard/CoffeeCard/Models/Product.cs
```diff
@@ -6,13 +6,20 @@ public class Product
...
-        [Required] public string Name { get; set; }
+        [Required]
+        public string Name { get; set; }
 
         public string Description { get; set; }
         public int ExperienceWorth { get; set; }
+        [Required]
+        public bool Visible { get; set; }
+        [Required]
+        public bool BaristasOnly { get; set; }
```
TTA777: We may want this as an enum, considering that different people has different prices, i.e board vs baristas and a few special people

coffeecard/CoffeeCard/Models/User.cs
```diff
@@ -30,8 +36,12 @@ public User()
         public virtual ICollection<Ticket> Tickets { get; set; }
         public virtual ICollection<Purchase> Purchases { get; set; }
         public virtual ICollection<Statistic> Statistics { get; set; }
+        [Required]
         public bool IsVerified { get; set; }
+        [Required]
         public bool PrivacyActivated { get; set; }
+        [Required]
+        public bool IsBarista { get; set; }
```
TTA777: Same as above, we may want to make it an enum, considering there is more than just baristas

coffeecard/CoffeeCard/Services/ProductService.cs
```diff
@@ -13,9 +15,26 @@ public ProductService(CoffeecardContext context)
...
         }
 
-        public IEnumerable<Product> GetProducts()
+        public async Task<IEnumerable<Product>> GetPublicProducts()
         {
-            return _context.Products.AsEnumerable();
+            return await GetProducts(false);
+        }
+
+        public async Task<IEnumerable<Product>> GetProductsForUserAsync(User user)
```
TTA777: If we are more specific than just barista, these would need to be changed ofcourse

jonasanker:
New commit https://github.com/AnalogIO/analog-core/pull/7/commits/22f426a1fd97d3c45fa9b96ab0aba69d9785a4f2, @TTA777. Addressing the wished changes.

coffeecard/CoffeeCard.Tests.Unit/Services/ProductServiceTest.cs
```diff
@@ -0,0 +1,369 @@
...
+                            NumberOfTickets = 10,
+                            Price = 10,
+                            ExperienceWorth = 10,
+                            Visible = true
+                        },
+                    };
+
+                    //var result = await productService.GetProductsForUserAsync(user);
+
+                    //Assert.Equal(expected, result);
```
TTA777: Why is this outcommented?

PR #11, opened 2020-03-02
Validate transaction amount before issuing a product

jonasanker:


coffeecard/CoffeeCard/Services/PurchaseService.cs
```diff
@@ -360,13 +359,24 @@ private void SendInvoiceEmail(Purchase purchase)
...
         {
             var purchase = _context.Purchases.FirstOrDefault(x => x.OrderId == payment.OrderId);
+
             if (purchase == null)
                 throw new ApiException($"The purchase with orderId {payment.OrderId} does not exist", 400);
+
             if (purchase.Completed) throw new ApiException("The given purchase has already been completed", 400);
 
-            var response = await _mobilePayService.GetPaymentStatus(payment.OrderId);
+            var mobilePayPaymentStatus = await _mobilePayService.GetPaymentStatus(payment.OrderId);
```
BBPL: Legit 💯
I think this is still exposed to security flaws, the original price should probably be taken from the Products, instead of the Purchase. The purchase can be faked, thus resulting in omitting the check!
MobilePay amount will always be equal to the purchase amount if tampered

jonasanker: The purchase is maintained by the server and is not exposed through controllers

BBPL: Ahhh yes, I have checked code in PurchaseService :) Looks good 👍

jonasanker: I do however agree with the point of making sure that the cannot be faked by any incoming request. 

A guess in this context the `Purchase` is the abstraction for a sales order and could in future development also be the one having knowledge about e.g. discounts or a sale consisting of  multiple products, so the price is told by purchase and not a product.

Regardless. Feel free to come with more feedback, if you stumble upon something :)

PR #13, opened 2020-03-05
Reworked the flow for recovering passwords

TTA777:
it now uses a website based on razor pages instead of sending an email with the password in plain text


coffeecard/CoffeeCard.WebApi/Models/Token.cs
```diff
@@ -18,5 +19,19 @@ public Token()
         public string TokenHash { get; set; }
 
         [ForeignKey("User_Id")] public virtual User User { get; set; }
+
+        public override bool Equals(object? obj)
```
jonasanker: How come object is nullable?

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
```diff
@@ -0,0 +1,56 @@
...
+            
+            if (await _accountService.RecoverUser(Token, PinCode.NewPinCode))
+            {
+                TempData["resultHeader"] = "Success";
+                TempData["result"] = "Your password has now been reset";
+            }
+            else
+            {
+                TempData["resultHeader"] = "Something went wrong";
+                TempData["result"] = "An error occured while updating your pin code. Please try again later";
```
jonasanker: Maybe ask the user to contact support@analogio.dk ?

coffeecard/CoffeeCard.WebApi/Services/TokenService.cs
```diff
@@ -35,20 +41,33 @@ public string GenerateToken(IEnumerable<Claim> claims)
...
         {
-            return new JwtSecurityTokenHandler().ReadToken(token) as JwtSecurityToken;
+            try
+            {
+                return new JwtSecurityTokenHandler().ReadToken(token) as JwtSecurityToken;
+            }
+            catch (Exception) //In case the string token is not in a JWS or JWE Compact Serialization Format 
+            {
+                return null;
+            }
```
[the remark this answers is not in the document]
jonasanker: Also log the exception here or somewhere else

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
```diff
@@ -42,6 +42,16 @@ public User GetAccountByEmail(string email)
...
+        private User GetAccountWithTokensByEmail(string email)
+        {
+            var user = _context.Users.
+                //Include(x => x.Programme).
+                Include(x => x.Tokens).
+                FirstOrDefault(x => x.Email == email);
+
+            if (user == null) throw new ApiException("No user found with the given email", 401);
+            return user;
+        }
```
jonasanker: Where do we use the token included in this method?

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
```diff
@@ -42,6 +42,16 @@ public User GetAccountByEmail(string email)
             if (user == null) throw new ApiException("No user found with the given email", 401);
             return user;
         }
+        private User GetAccountWithTokensByEmail(string email)
+        {
+            var user = _context.Users.
+                //Include(x => x.Programme).
```
jonasanker: pretty comment; remove uncommented line

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
```diff
@@ -217,37 +225,35 @@ public void ForgotPassword(string email)
...
             {
-                Log.Information("User tried to recover");
-                if (_tokenService.ValidateToken(tokenObj))
+                Log.Information($"User tried to recover with token {token}");
+                if (await _tokenService.ValidateToken(token))
                 {
                     var user = GetAccountByClaims(tokenObj.Claims);
-                    if (user != null)
+                    if (user != null) 
                     {
```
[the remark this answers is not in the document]
TTA777: Personally I was of the other opinion actually, but I don't have a strong opinion on it, so I'll invert them

coffeecard/CoffeeCard.WebApi/Services/TokenService.cs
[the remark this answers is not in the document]
TTA777: Would this be agreeable instead? Perhaps while logging that it received a malformed token if it can't read it?
```suggestion
            var jwtHandler = new JwtSecurityTokenHandler();
            if (jwtHandler.CanReadToken(token))
            {
                return new JwtSecurityTokenHandler().ReadToken(token) as JwtSecurityToken;
            }
            return null;
```

coffeecard/CoffeeCard.WebApi/Models/PagesModels/NewPinCodeModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace CoffeeCard.WebApi.Models.PagesModels
+{
+    public class NewPinCodeModel
+    {
+        [Required]
+        [RegularExpression(@"\d{4}", ErrorMessage = "The pin code must be four digits" )]
+        [DataType(DataType.Password, ErrorMessage = "The pin code must be four digits")]
```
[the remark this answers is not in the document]
TTA777: While testing it manually I was sometimes getting a generic error message when I didn't have it, though I can't seem to reproduce it now :/

coffeecard/CoffeeCard.WebApi/Models/Token.cs
TTA777: Dosen't it need to be in case you are trying to compare to null?

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
TTA777: This is the method that is used in ForgotPassword. The token is used in that one to check if there are currently any unused tokens for the user.

coffeecard/CoffeeCard.WebApi/Models/Token.cs
jonasanker: To my knowledge, the `if (obj is Token newToken)` is null safe

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
jonasanker: I cannot see where we do any check on the tokens?

coffeecard/CoffeeCard.WebApi/Services/TokenService.cs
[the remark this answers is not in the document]
jonasanker: My problem is more that you return null. It forces the rest of the code to do checks. I'll say just return the `return new JwtSecurityTokenHandler().ReadToken(token)`. Any exception thrown will just be propegated which would be on purpose :)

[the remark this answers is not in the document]
TTA777: Which forces me to handle the exception instead, or crash the website. I found it to be easier to just do a check for null, instead of having to handle the exception

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
TTA777: Ahh yes my mistake. The tokens included from that user is only used for reseting them just before we save the context. This is to mark them as used in our db. The place we are checking the token itself is validateToken in tokenService

coffeecard/CoffeeCard.WebApi/Models/Token.cs
TTA777: In that case, I can remove it I suppose

coffeecard/CoffeeCard.WebApi/Services/TokenService.cs
```diff
@@ -35,20 +41,33 @@ public string GenerateToken(IEnumerable<Claim> claims)
...
+                var tokenNotUsed = false;
+                var token = ReadToken(tokenString);
+
+                var user = await _claimsUtilities.ValidateAndReturnUserFromEmailClaimAsync(token.Claims);
+                
+                if (user.Tokens.Contains(new Token(tokenString))) tokenNotUsed = true;
+
+                return token.ValidTo > DateTime.UtcNow && tokenNotUsed;
             }
             catch (Exception)
```
jar-exerp: Try catch a specific exception

coffeecard/CoffeeCard.WebApi/Pages/Shared/_Layout.cshtml
```diff
@@ -21,13 +21,13 @@
         <p>
             <strong>Cafe Analog</strong><br/>
             Rued Langgaards Vej 7, 2300 Copenhagen S / CVR: 34657343<br/>
-            <a href="mailto:finance@cafeanalog.dk">finance@cafeanalog.dk</a> / <a href="https://www.cafeanalog.dk">www.cafeanalog.dk</a>
+            <a href="mailto:finance@cafeanalog.dk">support@analogio.dk</a> / <a href="https://www.cafeanalog.dk">www.cafeanalog.dk</a>
```
jar-exerp: Whoops, missing here :)

jar-exerp: See the mailto

coffeecard/CoffeeCard.WebApi/Services/TokenService.cs
```diff
@@ -36,21 +41,28 @@ public string GenerateToken(IEnumerable<Claim> claims)
...
+                var user = await _claimsUtilities.ValidateAndReturnUserFromEmailClaimAsync(token.Claims);
+                
+                if (user.Tokens.Contains(new Token(tokenString))) tokenNotUsed = true;
+
+                return token.ValidTo > DateTime.UtcNow && tokenNotUsed;
             }
-            catch (Exception)
+            catch (ArgumentException)
             {
                 return false;
```
jar-exerp: Maybe put a log here, to log what the exception was?
```suggestion
catch (ArgumentException e)
            {
                Log.LogError($"Unable to read toke. Exception thrown = {e}");
                return false;
```

jar-exerp:
Will the `baseUrl` be correct in this template? https://github.com/AnalogIO/analog-core/blob/feature/thomas/recover-site/coffeecard/CoffeeCard.WebApi/Services/EmailService.cs#L106

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
jar-exerp: I was just curious for if the method then is needed; can we clear the tokens without having included them in the first place? That was my thougt

jar-exerp:
We're are almost there, minor changes suggested. Else looks good!

You might want to consider to squash your commits before merging

coffeecard/CoffeeCard.WebApi/Startup.cs
```diff
@@ -163,6 +163,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             {
                 endpoints.MapControllers();
                 endpoints.MapRazorPages();
+                endpoints.MapFallbackToPage("/result");
```
jar-exerp: Does this mean that pages which does not match a routing will be routed to the result page?

coffeecard/CoffeeCard.WebApi/Pages/VerifyEmail.cshtml
```diff
@@ -0,0 +1,20 @@
...
+<html>
+<head>
+    <title></title>
+</head>
+<body>
+<div>
+    
+</div>
+</body>
+</html>
```
jar-exerp: These are empty, are they needed?

coffeecard/CoffeeCard.WebApi/Startup.cs
TTA777: Yes, and the fallback of the result page is to give an error saying this is not the page you are looking for.

coffeecard/CoffeeCard.WebApi/Pages/VerifyEmail.cshtml
TTA777: Probably not, since everything is happening in the model. I can try removing it if you consider it an eyesore

PR #18, opened 2020-05-02
New email templates and updated jquery

TTA777:
Updated three of the email templates to the new style. Update jquery to 3.5.0. Refactored EmailService.cs in order to avoid duplicate code.

I worked on this over quite a while in small sessions, so I may have forgotten about some changes, do double check


coffeecard/CoffeeCard.WebApi/wwwroot/Templates/EmailTemplate/email_verify_registration.html
```diff
@@ -1,356 +1,173 @@
...
+                            <tr>
+                                <td style="border-collapse: collapse !important; margin: 0 auto; mso-table-lspace: 0 !important; mso-table-rspace: 0 !important;cellpadding: 0; cellspacing: 0; border: 0; font-size: 1em">
+                                    <p style="margin: 0;>
+                                        You are almost there, you just need to verify that {email} is the email you want to use within the next {expiry}.
+                                    </p>
+                                    <!-- ======= start spacer ======= -->
+                                    <div style="font-size: 1px; height: 20px; line-height: 20px; Margin: 0; mso-line-height-rule: exactly;">&nbsp;</div>
+                                    <!-- ======= end spacer ======= -->
+                                    <p style="margin: 0;">
+                                        If you did not create an account with us you can safely disregard this message.
```
jonasanker, line 111: I think some welcome text is missing, could you add:
```suggestion
Thanks for registrering with the Cafe Analog App.

We need to verify your email before you can login. Click the button below to verify your email.

If you did not create an account with us, you can disregard this message.
```

coffeecard/CoffeeCard.WebApi/wwwroot/Templates/EmailTemplate/email_verify_updatedemail.html
```diff
@@ -1,356 +1,167 @@
...
+                    </td>
+                </tr>
+                <tr>
+                    <td style="border-collapse: collapse !important; margin: 0 auto; mso-table-lspace: 0 !important; mso-table-rspace: 0 !important;cellpadding: 0; cellspacing: 0; border: 0; align-content: center">
+                        <table border="0" cellpadding="0" cellspacing="0" align="center" width="90%" style="font-family: 'Nunito Sans', 'Helvetica Neue', helvetica, arial, sans-serif;  color:#38251a; align-content: center; text-align: center; margin: 0 auto; border-collapse: collapse !important; mso-table-lspace: 0 !important; mso-table-rspace: 0 !important;cellpadding: 0; cellspacing: 0; border: 0;">
+                            <tr>
+                                <td style="border-collapse: collapse !important; margin: 0 auto; mso-table-lspace: 0 !important; mso-table-rspace: 0 !important;cellpadding: 0; cellspacing: 0; border: 0; font-size: 1em">
+                                    <p style="margin: 0;">
+                                        You are almost there, you just need to verify your newly updated email for your account within {expiry} to be able to use it with the app.
```
jonasanker, line 105:
```suggestion
Your email was updated 

You are almost there..  Verify your updated email for your account within {expiry} to to use it with the Cafe Analog app.
```

coffeecard/CoffeeCard.WebApi/wwwroot/Templates/EmailTemplate/email_verify_registration.html
jonasanker: Maybe the user should contact us if someone registres with their email? Else it would cause trouble in the future. To be discussed

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
```diff
@@ -99,15 +91,10 @@ public void SendVerificationEmailForChangedEmail(User user, string token, string
         public void SendVerificationEmailForLostPw(User user, string token)
```
jonasanker, line 91: Rename to remove the verification part

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
```diff
@@ -77,16 +74,11 @@ public void SendRegistrationVerificationEmail(User user, string token)
         public void SendVerificationEmailForChangedEmail(User user, string token, string newEmail)
```
jonasanker, line 74: Consider outcommenting code and throw a `NotImplementedException`

jonasanker: Also we need to add a note to rethink the update email flow - should the email be updated before the new one is verified etc..

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
```diff
@@ -17,6 +17,7 @@ public class EmailService : IEmailService
         private const string MailgunApiUrl = "https://api.mailgun.net/v3";
```
jonasanker, line 17: Not on you, but it would be nice if this constant was moved to App Settings

jonasanker: `MailgunSettings` that would be

coffeecard/CoffeeCard.WebApi/wwwroot/Templates/EmailTemplate/email_verify_registration.html
TTA777: There was already the following as an intro message in the above paragraph: "_You are almost there, you just need to verify that {email} is the email you want to use within the next {expiry}._"

Your message might be a bit nicer, so I'll change the message to that. I don't think it would be a good idea to rely on the user to message us if they did not request the account, so if we are concerned about people using other emails to try and sign up I think we should take a different approach. I think it should be possible to do it using tokens so nothing is added to the database, except a token, before the email is validated.

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
TTA777: While easily done, I'm not sure I understand why you want to change it? It still serves to verify that the user requested a new password

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
TTA777: Added a todo and now throws the notImplemented exception as the first ting

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
TTA777: Moved there now. We'll need to remember updating the setting in our environments and in the vault

TTA777:
Build fails due to it not having the MailgunApiUrl in the appsettings I believe. It builds locally on my machine

jonasanker:
@TTA777  You are right about the cause of the failing test. Add the new field in the `appsetting-for-test.json` and the test should succeed

coffeecard/CoffeeCard.WebApi/Services/EmailService.cs
jonasanker: It's just hairsplitting, I'll approve the PR without this. My point was that I do not see the email as a verification email, more like a choose a new password email, hence we could remove the verification word :)

coffeecard/CoffeeCard.WebApi/wwwroot/Templates/EmailTemplate/email_verify_registration.html
jonasanker: That sounds like a good approach, could you create a small ticket on this on Trello?

PR #19, opened 2020-07-11
Implementation of login timeout that should be able to help alleviate potential problems with the login endpoint

TTA777:
As discussed earlier, here is a solution using a dictionary(map) in order to solve our potential issue

### DIFF
--- a/coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
+++ b/coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
@@ -11,10 +11,10 @@ namespace CoffeeCard.Common.Configuration
         [Required]
         public string AdminToken { get; set; }
         [Required]
-        public int MaximumLoginAttempts { get; set; }
+        public int MaximumLoginAttemptsWithinTimeOut { get; set; }
         [Required]
-        public TimeSpan TimeOut { get; set; }
-        
+        public int TimeOutPeriodInMinutes { get; set; }
+
         public void Validate()
         {
             Validator.ValidateObject(this, new ValidationContext(this), validateAllProperties: true);
--- a/coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
+++ b/coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.ComponentModel.DataAnnotations;
+﻿using System.ComponentModel.DataAnnotations;
 using NetEscapades.Configuration.Validation;
 
 namespace CoffeeCard.Common.Configuration
--- a/coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
+++ b/coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
@@ -1,17 +1,23 @@
-﻿using System.ComponentModel.DataAnnotations;
+﻿using System;
+using System.ComponentModel.DataAnnotations;
 using NetEscapades.Configuration.Validation;
 
 namespace CoffeeCard.Common.Configuration
 {
     public class IdentitySettings : IValidatable
     {
-        [Required] public string TokenKey { get; set; }
-
-        [Required] public string AdminToken { get; set; }
-
+        [Required]
+        public string TokenKey { get; set; }
+        [Required]
+        public string AdminToken { get; set; }
+        [Required]
+        public int MaximumLoginAttempts { get; set; }
+        [Required]
+        public TimeSpan TimeOut { get; set; }
+        
         public void Validate()
         {
-            Validator.ValidateObject(this, new ValidationContext(this), true);
+            Validator.ValidateObject(this, new ValidationContext(this), validateAllProperties: true);
         }
     }
 }
--- a/coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
+++ b/coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
@@ -60,7 +59,6 @@ namespace CoffeeCard.Tests.Unit.Services
                 SchemaName = "test"
             };
             var environmentSettings = new EnvironmentSettings();
-            var identitySettings = new IdentitySettings(){AdminToken = "test",MaximumLoginAttemptsWithinTimeOut = 5, TimeOutPeriodInMinutes = 1, TokenKey = "token"};
 
             var expectedResult = true;
             bool result;
--- a/coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
+++ b/coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
@@ -40,7 +40,7 @@ namespace CoffeeCard.Tests.Unit.Services
             {
                 var accountService = new AccountService(context, environmentSettings, new Mock<ITokenService>().Object,
                     new Mock<IEmailService>().Object, new Mock<IHashService>().Object,
-                    new Mock<IHttpContextAccessor>().Object, new Mock<ILoginLimiter>().Object, identitySettings );
+                    new Mock<IHttpContextAccessor>().Object, new Mock<ILoginLimiter>().Object);
                 result = await accountService.RecoverUser("bogus", "3433");
             }
 
@@ -60,7 +60,7 @@ namespace CoffeeCard.Tests.Unit.Services
                 SchemaName = "test"
             };
             var environmentSettings = new EnvironmentSettings();
-            var identitySettings = new IdentitySettings();
+            var identitySettings = new IdentitySettings(){AdminToken = "test",MaximumLoginAttemptsWithinTimeOut = 5, TimeOutPeriodInMinutes = 1, TokenKey = "token"};
 
             var expectedResult = true;
             bool result;
@@ -85,7 +85,7 @@ namespace CoffeeCard.Tests.Unit.Services
 
                 var accountService = new AccountService(context, environmentSettings, tokenService.Object,
                     new Mock<IEmailService>().Object, new Mock<IHashService>().Object,
-                    new Mock<IHttpContextAccessor>().Object, new Mock<ILoginLimiter>().Object, identitySettings );
+                    new Mock<IHttpContextAccessor>().Object, new Mock<ILoginLimiter>().Object);
                 
                 result = await accountService.RecoverUser("valid", "3433");
             }
@@ -106,7 +106,7 @@ namespace CoffeeCard.Tests.Unit.Services
                 SchemaName = "test"
             };
             var environmentSettings = new EnvironmentSettings();
-            var identitySettings = new IdentitySettings(){AdminToken = "test",MaximumLoginAttempts = 5, TimeOut = new TimeSpan(1,0,0), TokenKey = "token"};
+            var identitySettings = new IdentitySettings(){AdminToken = "test",MaximumLoginAttemptsWithinTimeOut = 5, TimeOutPeriodInMinutes = 5, TokenKey = "token"};
 
             var claim = new Claim(ClaimTypes.Email, "test@email.dk");
             var claims = new List<Claim> {claim};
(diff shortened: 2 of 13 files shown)


coffeecard/CoffeeCard.Common/Configuration/IdentitySettings.cs
```diff
@@ -1,17 +1,23 @@
...
-        [Required] public string AdminToken { get; set; }
-
+        [Required]
+        public string TokenKey { get; set; }
+        [Required]
+        public string AdminToken { get; set; }
+        [Required]
+        public int MaximumLoginAttempts { get; set; }
+        [Required]
+        public TimeSpan TimeOut { get; set; }
```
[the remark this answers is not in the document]
jonasanker: Then you could have a specialzed getter method returning a TimeSpan like:
```csharp
public TimeSpan getTimeOutAsTimeSpan()
{
     return new TimeSpan (hours: 0, minutes: TimeOut, seconds: 0);
}
```

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
```diff
@@ -142,5 +148,155 @@ await using (var context = new CoffeeCardContext(builder.Options, databaseSettin
...
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+
+                //Attempts to login a sixth time with the same credentials and captures the exception 
+                var tooManyLoginsException = Record.Exception(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                // Assert
+                Assert.NotNull(tooManyLoginsException);
+                Assert.IsType<ApiException>(tooManyLoginsException);
+                var tooManyLoginsApiException = (ApiException) tooManyLoginsException;
+                Assert.Equal(429,tooManyLoginsApiException.StatusCode); //Determines that the exception from the sixth attempt is due to too many logins
```
jonasanker, line 242: This would be cleaner
```suggestion
var expectedException = new ApiException($"Amount of failed login attempts exceeds the allowed amount, please wait for xx minutes", 429);
Assert.Equal(tooManyLoginsException, expectedException);
```

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
```diff
@@ -142,5 +148,155 @@ await using (var context = new CoffeeCardContext(builder.Options, databaseSettin
...
+                var accountService = new AccountService(context, environmentSettings, new Mock<ITokenService>().Object,
+                    new Mock<IEmailService>().Object, hasher.Object, httpContextAccessor.Object,
+                    new LoginLimiter(), identitySettings );
+
+                //Attempts to login 5 time with the wrong credentials 
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                Assert.Throws<ApiException>(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
```
jonasanker, line 234: Could also be done with a for-loop but that's a taste discussion.

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
```diff
@@ -142,5 +148,155 @@ await using (var context = new CoffeeCardContext(builder.Options, databaseSettin
...
+                    new Mock<IHttpContextAccessor>().Object, new Mock<ILoginLimiter>().Object, identitySettings );
+
+                actualToken = accountService.Login(user.Email, user.Password, "2.1.0");
+            }
+            
+            // Assert
+            Assert.Equal(expectedToken, actualToken);
+        }
+        
+        [Fact(DisplayName = "Login timeouts after five failed logins")]
```
jonasanker, line 195: It does not timeout, it rejects the login, right?

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
```diff
@@ -142,5 +148,155 @@ await using (var context = new CoffeeCardContext(builder.Options, databaseSettin
...
+                var tooManyLoginsException = Record.Exception(() => accountService.Login(user.Email, wrongPass, "2.1.0"));
+                // Assert
+                Assert.NotNull(tooManyLoginsException);
+                Assert.IsType<ApiException>(tooManyLoginsException);
+                var tooManyLoginsApiException = (ApiException) tooManyLoginsException;
+                Assert.Equal(429,tooManyLoginsApiException.StatusCode); //Determines that the exception from the sixth attempt is due to too many logins
+            }
+        }
+        
+               [Fact(DisplayName = "Login allows logins after timeout expired")]
```
jonasanker, line 246: Format issue

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
```diff
@@ -142,5 +148,155 @@ await using (var context = new CoffeeCardContext(builder.Options, databaseSettin
...
+            };
+            var environmentSettings = new EnvironmentSettings(){DeploymentUrl = "test", EnvironmentType = EnvironmentType.Test, MinAppVersion = "2.1.0"};
+            var identitySettings = new IdentitySettings(){AdminToken = "test",MaximumLoginAttempts = 5, TimeOut = new TimeSpan(1,0,0), TokenKey = "token"};
+
+            var userTokens = new List<Token>() {};
+            var user = new User {Id = 1, Name = "test", Tokens = userTokens, Email = "test@email.dk", Programme = new Programme(), Password = "test", IsVerified = true};
+
+            var wrongPass = "wrongPassword";
+            var hasher = new Mock<IHashService>();
+            hasher.Setup(h => h.Hash(wrongPass)).Returns(wrongPass);
```
jonasanker, line 214: You don't need this since you're testing a wrong password scenario, right?

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
```diff
@@ -142,5 +148,155 @@ await using (var context = new CoffeeCardContext(builder.Options, databaseSettin
...
+            // Arrange
+            var builder = new DbContextOptionsBuilder<CoffeeCardContext>()
+                .UseInMemoryDatabase(nameof(LoginGivenValidCredentialsReturnsToken));
+
+            var databaseSettings = new DatabaseSettings()
+            {
+                SchemaName = "test"
+            };
+            var environmentSettings = new EnvironmentSettings(){DeploymentUrl = "test", EnvironmentType = EnvironmentType.Test, MinAppVersion = "2.1.0"};
+            var identitySettings = new IdentitySettings(){AdminToken = "test",MaximumLoginAttempts = 5, TimeOut = new TimeSpan(1,0,0), TokenKey = "token"};
```
jonasanker, line 164: Formatting issue

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
```diff
@@ -40,9 +45,24 @@ public string Login(string username, string password, string version)
...
             {
+                var mapEntry = _loginLimiter.UpdateAndGetLoginAttemptCount(user.Email);
+
+                var loginAttemptsMade = mapEntry.Item2;
+                var timeSinceFirstFailedLogin = DateTime.UtcNow.Subtract(mapEntry.Item1);
+                if (loginAttemptsMade == maxAttempts && timeSinceFirstFailedLogin < timeout)
+                {
+                    Log.Warning("Login attempts exceeding maximum allowed for e-mail = {username} from IP = {ipadress} ", username,
+                        _httpContextAccessor.HttpContext.Connection.RemoteIpAddress);
+                    throw new ApiException($"Amount of failed login attempts exceeds the allowed amount, please wait for {timeout.TotalMinutes} minutes",
```
jonasanker, line 62: Consider throwing a more description exception instead of the generic ApiException

coffeecard/CoffeeCard.WebApi/Services/LoginLimiter.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+
+namespace CoffeeCard.WebApi.Services
+{
+    public class LoginLimiter : ILoginLimiter
```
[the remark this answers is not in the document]
jonasanker: One could also consider to make the method take a `User` as parameter and not a string. That would give future flexibility to change the map key. Could e.g. be the user id

coffeecard/CoffeeCard.WebApi/Services/AccountService.cs
TTA777: http 429 is for "too many connections" so thought it would be appropriate, any compelling reason not to use it?

coffeecard/CoffeeCard.WebApi/Services/LoginLimiter.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private ConcurrentDictionary<string, (DateTime, int)> _loginAttempts;
+
+        public LoginLimiter()
+        {
+            _loginAttempts = new ConcurrentDictionary<string, (DateTime, int)>();
+        }
+
+        /// <summary>
```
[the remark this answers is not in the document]
TTA777: Since the documentation in this instance is referring to the implementation I would think not? Otherwise, we are all of a sudden saying all potential implementations is implemented this way :/

coffeecard/CoffeeCard.Tests.Unit/Services/AccountServiceTest.cs
jonasanker: No it's fine

coffeecard/CoffeeCard.WebApi/Services/LoginLimiter.cs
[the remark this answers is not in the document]
jonasanker: I can see what you mean; let me rephrase :slightly_smiling_face: . For the most cases, the caller will use the interface and not call the implementation, so provide documentation for the caller, it would be most useful to add the documentation to the interface. However as you're also pointing out, your code comments address the implementation, so maybe a way to go about it, is to have the implementation specific comments on the implementation and the rest on the interface :slightly_smiling_face:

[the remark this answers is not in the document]
jonasanker: However, since the current level of documented methods, is really low, I would consider this change voluntary :blush:

coffeecard/CoffeeCard.WebApi/Services/LoginLimiter.cs
```diff
@@ -0,0 +1,71 @@
...
+            return _loginAttempts.AddOrUpdate(email, (DateTime.UtcNow, 0), (key, value) => (value.Item1, value.Item2 +1));
+        }
+
+        /// <summary>
+        /// Resets the first failed login time and in so doing, extends the lockout period
+        /// </summary>
+        /// <param name="email"></param>
+        private void ResetUsersTimeoutPeriod(string email)
+        {
+            if (!_loginAttempts.TryGetValue(email, out var oldEntry)) return;
```
jonasanker, line 40: You could refactor this and use the pair just retrieved from the map on line 52

coffeecard/CoffeeCard.WebApi/Services/LoginLimiter.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        /// <param name="user"></param>
+        /// <returns></returns>
+        public bool LoginAllowed(User user)
+        {
+            var (firstFailedLogin, loginAttemptsMade) = UpdateAndGetLoginAttemptCount(user.Email);
+            var timeOutPeriod = new TimeSpan(0, _identitySettings.TimeOutPeriodInMinutes, 0);
+            var timeSinceFirstFailedLogin = DateTime.UtcNow.Subtract(firstFailedLogin);
+
+
```
jonasanker, line 56: Load `5` from settings. Consider adding a comment to explain this expression

coffeecard/CoffeeCard.WebApi/Services/LoginLimiter.cs
TTA777: Since it is a map the operation should only cost O(1) right? I think the code is a bit cleaner this way, but if you disagree I'll reformat

Issue #41, opened 2021-11-09
Add userid to log


fremartini:
Investigate extending the log with userid. Ensure it is easy to extend with more attributes in the future
Possible implementation: https://github.com/ekmsystems/serilog-enrichers-correlation-id

PR #44, opened 2021-11-16
Feature/fremartini/add userid to log

fremartini:
closing #41


coffeecard/CoffeeCard.WebApi/Logging/Enrichers/UserIdEnricher.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            var userIdProperty = new LogEventProperty(UserIdPropertyName, new ScalarValue(GetUserId()));
+
+            logEvent.AddOrUpdateProperty(userIdProperty);
+        }
+
+        private string GetUserId()
+        {
+            var claims = _contextAccessor.HttpContext.User.Claims.ToList();
+            var id = claims.Count > 2 ? claims[2].Value : "NONE";
```
jonasanker, line 35: It seems a bit vulnerable to rely on a specific index to retrieve the UserId. I suggest using
```suggestion
            var id =  _contextAccessor.HttpContext.User.FindFirstValue(ClaimTypes.NameIdentifier);
```
See ref [What is the way in ASP.NET Core 3.0 to get current authenticated User username?](https://stackoverflow.com/questions/58473739/what-is-the-way-in-asp-net-core-3-0-to-get-current-authenticated-user-username)

coffeecard/CoffeeCard.WebApi/Logging/Enrichers/UserIdEnricher.cs
```diff
@@ -0,0 +1,40 @@
...
+            
+            return $"userid:{id}";
```
jonasanker, line 37: It would be better to not add the enricher in the log statement if no user is logged in, rather than adding `NONE`. Is that possible?

coffeecard/CoffeeCard.WebApi/Logging/Enrichers/UserIdEnricher.cs
```diff
@@ -0,0 +1,40 @@
...
+        private readonly IHttpContextAccessor _contextAccessor;
+
+        public UserIdEnricher() : this(new HttpContextAccessor())
+        {
+        }
+
+        private UserIdEnricher(IHttpContextAccessor contextAccessor)
+        {
+            _contextAccessor = contextAccessor;
+        }
```
jonasanker, line 20: Do you know if we can just use Dependency Injection? I can see you are just following the pattern from `CorrelationIdEnricher` but wondering if below shouldn't be possible:

```suggestion
        public UserIdEnricher(IHttpContextAccessor contextAccessor)
        {
            _contextAccessor = contextAccessor;
        }
```

jonasanker (changes requested):
Overall looks good. I have added some changes I think you should address

fremartini:
@jonasanker considered and implemented proposed changes

jonasanker (approved):
Approved with latest changes

PR #52, opened 2022-01-06
Fix GetSemesterStart

marfavi:
Fix a bug that would attempt to return a DateTime with day > 31
Add guard clause for autumn semester case

### DIFF
--- a/coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
+++ b/coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
@@ -16,7 +16,7 @@ namespace CoffeeCard.Tests.Unit.Models
         [InlineData("2030/06/30", "2030/01/28")]
         [InlineData("2000/07/01", "2000/07/01")]
         [InlineData("2020/12/31", "2020/07/01")]
-        public void TestSemesterStartGetsLastMondayOfJanuaryInSpring(string currentDateStr, string expectedDateStr)
+        public void TestGetSemesterStart(string currentDateStr, string expectedDateStr)
         {
             var currentDate = DateTime.Parse(currentDateStr);
             var expectedDate = DateTime.Parse(expectedDateStr);
--- a/coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
+++ b/coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
@@ -16,11 +16,8 @@ namespace CoffeeCard.Tests.Unit.Models
         [InlineData("2030/06/30", "2030/01/28")]
         public void TestSemesterStartGetsLastMondayOfJanuaryInSpring(string currentDateStr, string expectedDateStr)
         {
-            var cd = currentDateStr.Split('/');
-            var ed = expectedDateStr.Split('/');
-
-            var currentDate = new DateTime(int.Parse(cd[0]), int.Parse(cd[1]), int.Parse(cd[2]));
-            var expectedDate = new DateTime(int.Parse(ed[0]), int.Parse(ed[1]), int.Parse(ed[2]));
+            var currentDate = DateTime.Parse(currentDateStr);
+            var expectedDate = DateTime.Parse(expectedDateStr);
 
             var actualDate = Statistic.GetSemesterStart(currentDate);
             Assert.Equal(expectedDate, actualDate);
--- a/coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
+++ b/coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
@@ -6,7 +6,7 @@ namespace CoffeeCard.Tests.Unit.Models
 {
     public class SemesterStartTests
     {
-        [Theory(DisplayName = "When called in months Jan-Jun (inclusive), GetSemesterStart returns the last Monday of January")]
+        [Theory(DisplayName = "GetSemesterStart returns the correct semester start date based on currentDate")]
         [InlineData("2022/01/31", "2022/01/31")]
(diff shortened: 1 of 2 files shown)


coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
```diff
@@ -0,0 +1,29 @@
...
+        [InlineData("2026/01/01", "2026/01/26")]
+        [InlineData("2027/01/01", "2027/01/25")]
+        [InlineData("2030/06/30", "2030/01/28")]
+        public void TestSemesterStartGetsLastMondayOfJanuaryInSpring(string currentDateStr, string expectedDateStr)
+        {
+            var cd = currentDateStr.Split('/');
+            var ed = expectedDateStr.Split('/');
+
+            var currentDate = new DateTime(int.Parse(cd[0]), int.Parse(cd[1]), int.Parse(cd[2]));
+            var expectedDate = new DateTime(int.Parse(ed[0]), int.Parse(ed[1]), int.Parse(ed[2]));
```
jonasanker, line 23: We can take use of the `DateTime.Parse()` method instead of parsing it manually. Also makes the code more clean
```suggestion
            var currentDate = DateTime.Parse(currentDateStr);
            var expectedDate = DateTime.Parse(expectedDateStr);
```

coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class SemesterStartTests
+    {
+        [Theory(DisplayName = "When called in months Jan-Jun (inclusive), GetSemesterStart returns the last Monday of January")]
+        [InlineData("2022/01/31", "2022/01/31")]
+        [InlineData("2023/01/31", "2023/01/30")]
+        [InlineData("2024/05/12", "2024/01/29")]
+        [InlineData("2025/06/24", "2025/01/27")]
+        [InlineData("2026/01/01", "2026/01/26")]
+        [InlineData("2027/01/01", "2027/01/25")]
+        [InlineData("2030/06/30", "2030/01/28")]
```
jonasanker, line 16: Since we are testing the method, it would be great to create a separate test for the case of the autumn semester.

coffeecard/CoffeeCard.WebApi/Models/Statistic.cs
```diff
@@ -25,16 +25,13 @@ public class Statistic
 
         public static DateTime GetSemesterStart(DateTime currentTime)
```
jonasanker, line 26: This is not on you but these methods do not make sense to have on a model class; `Statistics`. It would be much more meaning full to move these utility methods to a separate static helper class, 

Also to make it more clear for future users, we should probably add some documentation. I've added a suggestion

```suggestion
        /// <summary>
        /// Get the start of date of the semester.
        /// When <paramref name="currentTime"/> is in July (Month 7) or greater, the semester start is July 1st or when earlier than July, the semester start is the date of the last Monday of January.
        /// </summary>
        public static DateTime GetSemesterStart(DateTime currentTime)
```

coffeecard/CoffeeCard.WebApi/Models/Statistic.cs
```diff
@@ -25,16 +25,13 @@ public class Statistic
...
+            if (currentTime.Month >= 7) return new DateTime(currentTime.Year, 7, 1);
 
-            return new DateTime(currentTime.Year, 7, 1);
+            // Spring semester: Get last Monday of January.
+            var lastDayOfJan = new DateTime(currentTime.Year, 1, 31);
+            int correctedLastDayOfWeek = (int)(lastDayOfJan.DayOfWeek + 6) % 7; // Mon=0, Tue=1, ..., Sun=6
+            return lastDayOfJan.AddDays(-correctedLastDayOfWeek);
         }
 
         public static DateTime GetSemesterEnd(DateTime currentTime)
```
jonasanker, line 37: Let's add some documentation while we're here. We could also unit test this one but will not require this of you 😉 

```suggestion
        /// <summary>
        /// Get the end of date of the semester.
        /// When <paramref name="currentTime"/> is in July (Month 7) or greater, the semester start is December 23rd or when earlier than July, the semester start is June 30rd.
        /// </summary>
        public static DateTime GetSemesterEnd(DateTime currentTime)
```

jonasanker (changes requested):
A few minor changes suggested

coffeecard/CoffeeCard.WebApi/Models/Statistic.cs
marfavi: > This is not on you but these methods do not make sense to have in a model class; `Statistics`. It would be much more meaning full to move these utility methods to a separate static helper class,

I agree. This should be added as an issue.

coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
marfavi: Added and changed the display name of the theory.

coffeecard/CoffeeCard.Tests.Unit/Models/SemesterStartTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        [InlineData("2000/07/01", "2000/07/01")]
+        [InlineData("2020/12/31", "2020/07/01")]
+        public void TestSemesterStartGetsLastMondayOfJanuaryInSpring(string currentDateStr, string expectedDateStr)
```
jonasanker, line 19: The name of the method is not accurate with the addition of the autumn semester test cases. 

```suggestion
        public void TestGetSemesterStart(string currentDateStr, string expectedDateStr)
```

jonasanker (approved):
One minor and then we're good to go 👍🏻

PR #55, opened 2022-01-08
Users may only retrieve their own info

TTA777:
Added a check, so an authenticated user may only request their own information


jonasanker (approved):
LGTM

Issue #42, opened 2021-11-09
Update the swagger doc with the return type on endpoints


TTA777:
Currently swagger does not document the refsponse type of the different endpoints. This can be added by adding by using sometihng similar to this for each endpoint:

 [ProducesResponseType(typeof(IsOpenDTO), 200)]

PR #57, opened 2022-01-09
OpenApi specification and documentation

jonasanker:
- Upgraded from Swagger to OpenApi v3
- Documented all endpoints (controllers) and their request/response Dtos
- Refactored project structure with the new `CoffeeCard.Library` project. This is to separate the service and persistence layer from the WebApi layer
- OpenApi/SwaggerUi document improvements
- Specify Bearer token authentication
- Generate an Api specification for each Api version
- Custom Api processor
- Bump NuGet packages and fix EmailService after package upgrade 

Closing #42


jonasanker:
I wonder if the Models should be moved to either `CoffeeCard.Library` or a new `CoffeeCard.WebApi`?

I don't think they belong to the Common project by a second thought. The Common project should be the most minimal project with classes all projects have in common.

TTA777 (approved):
Lgtm, though it seems the pipeline needs to be updated with the new location of www.analogio.dk.pfx and appsettings.json in the mobilepay client project

PR #66, opened 2022-01-13
Generated MobilePay Payments and Webhook API client

jonasanker:
The APIs are generated by running the Main method in the CoffeeCard.MobilePay.GenerateApi project


jonasanker:
Maybe `MobilePay.GenerateApi` should be moved to a more generic program in case we want do the same with other APIs we consume.

PR #74, opened 2022-02-05
Use Postgres

jonasanker:
We move to Postgres DB when moving to Digital Ocean

- EF versions bumped
- Use Postgres SQL driver
- Update migrations


fremartini (approved):
LGTM. What is the motivation for using Postgres instead of what we used before?

jonasanker:
@fremartini Digital Ocean does not offer Microsoft SQL Server but does support Postgres SQL which is a highly renowned open-source database. So this PR is about changing to the Postgres driver

PR #76, opened 2022-03-06
delete account endpoint

### LINKED ISSUE #68: New Delete Account endpoint
Introduce a new endpoint for deleting an account.

Suggested API
`DELETE api/v2/account`
- 204 No content upon successful request
- 404 Account not found

Tasks
- All user information on the user entry should be overwritten with anonymized data, so that purchase history etc are not deleted but also not referable to a _person_
- The anonymized ("deleted user") should not be visible on leaderboards
- A user should verify his intent to delete the account, it could e.g. be done in a similar way as we do when verifying an email address
- (If possible) User's JWT token should be invalidated

Related to https://github.com/AnalogIO/coffeecard_app/issues/72

fremartini:
closes #68


fremartini:
@TTA777 how do you test the mailgun logic?

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <param name="email">email of account to delete</param>
+        /// <response code="204">Successful account deletion</response>
+        /// <response code="404">Account not found</response>
+        /// <response code="401">Invalid credentials</response>
+        [HttpDelete("account")]
+        [ProducesResponseType(typeof(void), StatusCodes.Status204NoContent)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiException), StatusCodes.Status429TooManyRequests)]
+        public ActionResult Delete([FromBody] string email)
```
jonasanker, line 41: There is no verification of the user token is the same as the user who owns the email. This would allow all users to request account deletion for all other users. We should identify the user from the token and then request the account deletion for that specific user.
The `ClaimsUtilities.ValidateAndReturnUserFromClaimAsync` should do the job for you, then you can just parse the user class as parameter to the `AnonymizeAccount` method

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace CoffeeCard.WebApi.Controllers.v2
+{
+    /// <summary>
+    /// Controller for creating and managing user accounts
+    /// </summary>
+    [ApiController]
+    [ApiVersion("2")]
+    [Route("api/v{version:apiVersion}/[controller]")]
```
jonasanker, line 14: We want to force the url for this controller to be in lowercase
```suggestion
    [Route("api/v{version:apiVersion}/account")]
```

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        /// <summary>
+        /// Delete an account
+        /// </summary>
+        /// <param name="email">email of account to delete</param>
+        /// <response code="204">Successful account deletion</response>
+        /// <response code="404">Account not found</response>
+        /// <response code="401">Invalid credentials</response>
+        [HttpDelete("account")]
```
jonasanker, line 36: Adding `account` to `HttpDelete` appends the account to the route, which will give the route `/api/v2/Account/account`. I think it would be more RESTful with just `DELETE /api/v2/Account`.

```suggestion
        [HttpDelete]
```

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+        [ProducesResponseType(typeof(void), StatusCodes.Status204NoContent)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status404NotFound)]
```
jonasanker, line 38: With the change to only delete the account of a logged-in user, I'm not sure we can hit a 404 scenario

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +248,60 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
...
+            var user = _context.Users.First(x => x.Email == email);
+            
+            var claims = new[]
+            {
+                new Claim(ClaimTypes.Email, user.Email), new Claim(ClaimTypes.Name, user.Name),
+                new Claim(ClaimTypes.Role, "verification_token")
+            };
+            var verificationToken = _tokenService.GenerateToken(claims);
+            
+            _emailService.SendVerificationEmailForDeleteAccount(user, verificationToken);
```
jonasanker, line 268: This is an async method. Should be awaited

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +248,60 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
...
+        }
+        private void AnonymizeUser(User user)
+        {
+            user.Email = string.Empty;
+            user.Name = string.Empty;
+            user.Password = string.Empty;
+            user.Salt = string.Empty;
+            user.DateUpdated = DateTime.Now;
+            user.PrivacyActivated = true;
+            _context.SaveChanges();
```
jonasanker, line 288: Use the async equivalent

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +248,60 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
...
+        }
+        
+        public bool VerifyAnonymization(string token)
+        {
+            Log.Information($"Trying to verify deletion with token: {token}");
+
+            var user = VerifyTokenClaimAndUser(token);
+
+            AnonymizeUser(user);
+            return _context.SaveChanges() > 0;
```
jonasanker, line 278: Will this not always return false since we already saved the changes in `AnonymizeUser`? Also, I think returning `void` is just fine since an exception should be thrown if we can not save the user etc.

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +248,60 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
             await _context.SaveChangesAsync();
             return true;
         }
+        
+        public bool UserExists(string email)
+        {
+            return _context.Users.Any(x => x.Email == email);
```
jonasanker, line 254: Async

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +248,60 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
...
+            {
+                new Claim(ClaimTypes.Email, user.Email), new Claim(ClaimTypes.Name, user.Name),
+                new Claim(ClaimTypes.Role, "verification_token")
+            };
+            var verificationToken = _tokenService.GenerateToken(claims);
+            
+            _emailService.SendVerificationEmailForDeleteAccount(user, verificationToken);
+        }
+        
+        public bool VerifyAnonymization(string token)
```
jonasanker, line 271: I think the VerifyAnonymization method name is a bit misleading since we do not verify anonymization, but instead, _performs_ the anonymization. From the client, it is seen as a verification but I think method name should reflect what it actually performs

jonasanker (changes requested):
A few comments. Most critically is the need to fix that a user can only delete his own account and not everyone's 

We discussed introducing a new user state. Will you add the code for this? This will also allow us for rejecting a user using his token after the account deletion

coffeecard/CoffeeCard.Library/Services/IAccountService.cs
```diff
@@ -16,5 +16,8 @@ public interface IAccountService
         void UpdateExperience(int userId, int exp);
         Task ForgotPasswordAsync(string email);
         Task<bool> RecoverUserAsync(string token, string newPassword);
+        bool UserExists(string email);
+        void AnonymizeAccount(string email);
+        bool VerifyAnonymization(string token);
```
jonasanker, line 21: These should all be async since they involve contacting the DB layer

coffeecard/CoffeeCard.Library/Services/AccountService.cs
jonasanker: The same would also apply to the name of the method called from the controller. It should have a name reflecting that we are making a request to delete an account but doesn't do it in that method call

TTA777:
@fremartini The reason we want to catch any exceptions here are that the user will be presented with plain json saying the an unhandled error occurred otherwise, since we are on a website now, and not using a http client. See here if you want to [test it](https://api-test.analogio.dk/coffeecard/verifyEmail?token=test). We just hadn't run into the bug before, and it is an unlikely scenario, but nonetheless one that should be fixed, as you can see. Please do fix it on the verification page as well

Regarding mailgun, it will likely just work, it seems you are using the existing code just fine there. What has been a pain point in the past is how email clients display the email when they receive it. It appears you are making use of the mjml framework that I introduced at some point in the past to resolve this issue, so it likely won't be an issue either. But do go through the process once you have it in the test environment.

fremartini:
> A few comments. Most critically is the need to fix that a user can only delete his own account and not everyone's
> 
> We discussed introducing a new user state. Will you add the code for this? This will also allow us for rejecting a user using his token after the account deletion

So you are suggesting adding an extra field 'Status' that is either
* Active
* Deleted
*Pending

correct?

jonasanker:
> > A few comments. Most critically is the need to fix that a user can only delete his own account and not everyone's
> > We discussed introducing a new user state. Will you add the code for this? This will also allow us for rejecting a user using his token after the account deletion
> 
> So you are suggesting adding an extra field 'Status' that is either
> 
> * Active
> * Deleted
>   *Pending
> 
> correct?

Yes, you got it. I think maybe `Pending` is too generic a name, maybe `Pending_Email_Verification` is more descriptive? Just a thought

fremartini:
> 

so how would i do this? i assume this requires a change in the data model. Maybe we can have a look at this sunday?

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            _accountService = accountService;
+            _claimsUtilities = claimsUtilities;
+        }
+
+        /// <summary>
+        /// Delete an account
+        /// </summary>
+        /// <param name="email">email of account to delete</param>
+        /// <response code="204">Successful account deletion</response>
```
jonasanker, line 37: The account is not deleted when 204 is returned. We should to the documentation in the summary section that an anonymization email is sent etc and the 204 response represent a successful initiation of the anonymization request

coffeecard/CoffeeCard.WebApi/Pages/VerifyEmail.cshtml.cs
```diff
@@ -17,17 +18,24 @@ public VerifyEmail(IAccountService accountService)
...
+                    return RedirectToPage("result");
+                }
             }
-
-            TempData["resultHeader"] = "Error";
-            TempData["result"] =
-                @"Looks like the link you used has expired or already been used. Request a new password in the app to verify your email.";
+            catch (Exception)
+            {
+                //FIXME: how do we handle this error? display it? log it?
```
jonasanker, line 33: What error are we expecting, what are we trying to catch here?

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+        public AccountController(IAccountService accountService, ClaimsUtilities claimsUtilities)
+        {
+            _accountService = accountService;
+            _claimsUtilities = claimsUtilities;
+        }
+
+        /// <summary>
+        /// Delete an account
+        /// </summary>
+        /// <param name="email">email of account to delete</param>
```
jonasanker, line 36: We do not use this parameter anymore

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Initializes a new instance of the <see cref="AccountController"/> class.
+        /// </summary>
+        public AccountController(IAccountService accountService, ClaimsUtilities claimsUtilities)
+        {
+            _accountService = accountService;
+            _claimsUtilities = claimsUtilities;
+        }
+
+        /// <summary>
+        /// Delete an account
```
jonasanker, line 34: Add a documentation on what is happening

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -0,0 +1,52 @@
...
+    [ApiController]
+    [ApiVersion("2")]
+    [Route("api/v{version:apiVersion}/account")]
+    [Authorize]
+    public class AccountController : ControllerBase
+    {
+        private readonly IAccountService _accountService;
+        private readonly ClaimsUtilities _claimsUtilities;
+        
+
```
jonasanker, line 23: Too much space

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +254,52 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
...
+            {
+                new Claim(ClaimTypes.Email, user.Email), new Claim(ClaimTypes.Name, user.Name),
+                new Claim(ClaimTypes.Role, "verification_token")
+            };
+            var verificationToken = _tokenService.GenerateToken(claims);
+            
+            await _emailService.SendVerificationEmailForDeleteAccount(user, verificationToken);
+        }
+        
+        public void AnonymizeAccount(string token)
```
jonasanker, line 269: Shouldn't this be async since e.g. the AnonymizeUser method call is async

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -256,6 +254,52 @@ public async Task<bool> RecoverUserAsync(string token, string newPassword)
...
+        
+        public async void AnonymizeAccount(string token)
+        {
+            Log.Information($"Trying to verify deletion with token: {token}");
+
+            var user = VerifyTokenClaimAndUser(token);
+
+            await AnonymizeUser(user);
+        }
+        private async Task AnonymizeUser(User user)
```
jonasanker, line 277: No space between methods 😊

coffeecard/CoffeeCard.Models/Entities/User.cs
```diff
@@ -144,4 +148,12 @@ public void IncreaseStatisticsBy(int increaseBy)
             monthStatistics.LastSwipe = utcNow;
         }
     }
+    
+    public enum  UserState
+    {
+        Active,
+        Deleted,
+        PendingActivition,
+
```
jonasanker, line 157: Why the comma and space?

jonasanker (approved):
Minor comments

PR #94, opened 2022-08-31
Migrate LeaderboardController to API v2

### LINKED ISSUE #93: Include UserId in Leaderboard endpoint

jonasanker:
- GET Leaderboard returns list of Leaderboard entries with Account Id included
- Fix bug #95 
- Extract methods from Entity classes

Closes #93, #95


TTA777 (changes requested):
I might be missing something, but how does this address the current problem we have in the app?

As I understand it, we want to show the user's score when they are outside the top 10. How will you accomplish this, by simply including the IDs of the top 10?

Don't we want to either create a new endpoint where we can get the score of a user by its ID instead, or expand the user endpoint to include said information?

jonasanker:
> I might be missing something, but how does this address the current problem we have in the app?
> As I understand it, we want to show the user's score when they are outside the top 10. How will you accomplish this, by simply including the IDs of the top 10?

Good question. Per the discussion on Slack, the leaderboard will temporarily only show the top 10. This update allows the app to detect if the logged in user is on the Top 10 list.

> Don't we want to either create a new endpoint where we can get the score of a user by its ID instead, or expand the user endpoint to include said information?

We might go here in the future. The GET Account endpoint already provides the rank, so that is one path to go. The alternative as you mention is to introduce a separate endpoint. We should create an issue with the solution we want once discussed.

TTA777:
> This update allows the app to detect if the logged in user is on the Top 10 list.
> 
Since we can already determine this by the rank returned by the GET Account endpoint, what does this PR offer us, that we are currently unable to do?

jonasanker:
> > This update allows the app to detect if the logged in user is on the Top 10 list.
> > 
> Since we can already determine this by the rank returned by the GET Account endpoint, what does this PR offer us, that we are currently unable to do?
> 

Before the Account Id was not included in the list of leaderboard entries. We need the Account Id to compare the leaderboard users so you can check if one of them is the currently logged in user.

TTA777:
> > > This update allows the app to detect if the logged in user is on the Top 10 list.
> > 
> > 
> > Since we can already determine this by the rank returned by the GET Account endpoint, what does this PR offer us, that we are currently unable to do?
> 
> Before the Account Id was not included in the list of leaderboard entries. We need the Account Id to compare the leaderboard users so you can check if one of them is the currently logged in user.

Since we are already determining if the current user is in the top 10 when we highlight them on the leaderboard, I don't understand what feature this enables us to do in the app. Is the motivation simply to make that check more reliable, due to the slightly hacky way we currently do it?

TTA777 (approved):
Lgtm, nice that we get a few more tests.

PR #96, opened 2022-09-04
Introduce new Leaderboard endpoint for retrieving rank and score of a…

jonasanker:
…uthenticated user

- Refactor usage of SemesterUtils
- Improve test coverage

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
@@ -28,7 +28,7 @@ namespace CoffeeCard.Library.Services.v2
                 .ToListAsync();
 
             var topStatistics = allStatisticsByPreset
-                .Where(s => ParsePresetAndCheckValidity(s, preset))
+                .Where(IsSwipeValid)
                 .OrderByDescending(s => s.SwipeCount)
                 .ThenBy(s => s.LastSwipe)
                 .Take(top);
@@ -49,7 +49,7 @@ namespace CoffeeCard.Library.Services.v2
                 .ToListAsync();
 
             var sortedStatistics = allStatisticsByPreset
-                .Where(s => ParsePresetAndCheckValidity(s, preset))
+                .Where(IsSwipeValid)
                 .OrderByDescending(s => s.SwipeCount)
                 .ThenBy(s => s.LastSwipe)
                 .ToList();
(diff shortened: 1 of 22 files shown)


coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
```diff
@@ -1,37 +1,92 @@
...
+            var now = _dateTimeProvider.UtcNow();
+            
+            return preset switch
+            {
+                LeaderboardPreset.Total => s.Preset == StatisticPreset.Total,
+                LeaderboardPreset.Semester => s.Preset == StatisticPreset.Semester &&
+                                              SemesterUtils.IsSwipeValidInSemester(s.LastSwipe, now),
+                LeaderboardPreset.Month => s.Preset == StatisticPreset.Monthly &&
+                                           SemesterUtils.IsSwipeValidInMonth(s.LastSwipe, now),
+                _ => true
```
marfavi, line 88: Should the default case be `false`, in case we add any more `LeaderboardPreset`s in the future and forget to add the case to this switch?

coffeecard/CoffeeCard.Library/Utils/DateTimeProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
marfavi, line 1: What is the motivation of this class? I'm thinking it's used to setup DI?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Leaderboard/LeaderboardEntry.cs
```diff
@@ -26,11 +29,36 @@ public class LeaderboardEntry
...
+        public int Rank { get; set; }
+        
         /// <summary>
         /// Account score
         /// </summary>
         /// <value>Account score</value>
         /// <example>25</example>
         public int Score { get; set; }
+
+        protected bool Equals(LeaderboardEntry other)
```
marfavi, line 46: Should Equals and GetHashCode have doc comments? I'm getting the following warnings:

Missing XML comment for publicly visible type or member 'LeaderboardEntry.Equals(object)' [CoffeeCard.Models]csharp(CS1591)

marfavi (commented):
Looks good. I have a few clarifying questions

coffeecard/CoffeeCard.Library/Utils/DateTimeProvider.cs
jonasanker: Calling `DateTime.UtcNow` is a side effect. For testing we need to wrap it, so we can mock the DateTime

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Leaderboard/LeaderboardEntry.cs
jonasanker: True that it gives a warning, however I don't think there is much value in documenting a vanilla method, so I have disabled the warning in this class

coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
jonasanker: I get your point. While not having the strongest opinion, my thought is that we here look at what presets have a restriction on the validity. This is only the case for Semester and Monthly and not the default case of Total. I have refactored the method a bit to reflect this more clearly

marfavi (approved):

PR #102, opened 2022-10-13
Include User data when querying Leaderboard Statistics

jonasanker:
Closes #101

PR #103, opened 2022-10-19
Merge to Production

jonasanker:
- Include User data when querying Leaderboard Statistics (#102)


TTA777 (commented):
So the new docker configuration for development does not seem to work on my machine, though this could be because I am using redhat linux, which has a few quirks every once in a while. You have verified it to work for you I take it?

jonasanker:
@TTA777 I believe both Frederik and I got the image to work. What error do you get?

TTA777:
@jonasanker Looked into it a bit more, and turns out it was SeLinux causing the issue. I've got it working now

jonasanker:
@TTA777 The branch not being up to date is not due to content but merge commits. I think you can go ahead and approve

TTA777 (approved):
lgtm

Issue #91, opened 2022-06-01
Register user with set programme


marfavi:
The register user DTO should have a required field `programme`. User will be assigned to that programme on registration instead of assigning the user the default `SWU` programme

Unblocks https://github.com/AnalogIO/coffeecard_app/issues/124

jonasanker:
Obs. Adding a new required field is a breaking change so this change must introduced in API v2

PR #106, opened 2022-10-25
Add v2 api for registering with programme

Hjaltesorgenfrei:
solves #91


jonasanker:
Sorry for being a bit slow here. I will review this on Sunday

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -31,6 +32,27 @@ public AccountController(IAccountService accountService, ClaimsUtilities claimsU
             _claimsUtilities = claimsUtilities;
         }
 
+                /// <summary>
```
jonasanker, line 35: Indentation is off

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -31,6 +32,27 @@ public AccountController(IAccountService accountService, ClaimsUtilities claimsU
...
+        /// <response code="409">Email already registered</response>
+        [HttpPost("register")]
+        [AllowAnonymous]
+        [ProducesResponseType(typeof(MessageResponseDto), StatusCodes.Status201Created)]
+        [ProducesResponseType(typeof(MessageResponseDto), StatusCodes.Status409Conflict)]
+        public async Task<ActionResult<MessageResponseDto>> Register([FromBody] RegisterDto registerDto)
+        {
+            await _accountService.RegisterAccountAsync(registerDto.Name, registerDto.Email, registerDto.Password, registerDto.ProgrammeId);
+
+            return Created("get", new MessageResponseDto()
```
jonasanker, line 49: Have you verified that the `get` URI is the correct one? Since this is the V2 API, we're currently relying on the V1 for GET Account. 
Also, a would prefer a static reference to the URI if possible. Normally that can be achieved when actions are in the same Controller but not sure what is possible here, since you are relying on the GET from V1

jonasanker (changes requested):
I have a few comments I would like to address, see the specific classes.

In the most recent V2 API implementations, I have used this as an opportunity to rewrite the service layers as some of these are a bit messy. That means for a new Controller, I have also created a new corresponding V2 service which has been rewritten to a better standard, and more importantly, we have added tests to it.
We did not follow this strictly for the first work in the AccountController V2, so I will not force this on you but it would be highly appreciated if we could do this for Account service too.

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
Hjaltesorgenfrei: I think I hardcoded the path to the v1

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -94,31 +94,29 @@ public string Login(string email, string password, string version)
...
 
             var salt = _hashService.GenerateSalt();
-            var hashedPassword = _hashService.Hash(registerDto.Password + salt);
+            var hashedPassword = _hashService.Hash(password + salt);
 
-            //This can potentially be implemented again, but is just sat to 1 for now
-            var programme = _context.Programmes.FirstOrDefault(x => x.Id == 1);
-            if (programme == null) throw new ApiException("No programme found with the id: 0", StatusCodes.Status400BadRequest);
+            var chosenProgramme = _context.Programmes.FirstOrDefault(x => x.Id == programme);
+            if (chosenProgramme == null) throw new ApiException("No programme found with the id: 0", StatusCodes.Status400BadRequest);
```
jonasanker, line 111: This should be the value of the provided programmeId?

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -94,31 +94,29 @@ public string Login(string email, string password, string version)
                 StatusCodes.Status401Unauthorized);
         }
 
-        public async Task<User> RegisterAccountAsync(RegisterDto registerDto)
+        public async Task<User> RegisterAccountAsync(string name, string email, string password, int programme = 1)
```
jonasanker, line 97: Document what programme = 1 means, e.g. SWU. However, does it make sense to have a default value, when is that the case?

jonasanker: Ahh, I suppose for backwards compatibility?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/RegisterDTO.cs
```diff
@@ -0,0 +1,51 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace CoffeeCard.Models.DataTransferObjects.V2.User
+{
+    /// <summary>
+    /// Register a new user data object
```
jonasanker, line 6: Register a new user?

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -31,6 +32,27 @@ public AccountController(IAccountService accountService, ClaimsUtilities claimsU
             _claimsUtilities = claimsUtilities;
         }
 
+        /// <summary>
+        /// Register a new account. A account is required to verify its email before logging in
```
jonasanker, line 36: Grammar: An account..

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -31,6 +32,27 @@ public AccountController(IAccountService accountService, ClaimsUtilities claimsU
...
+        /// </summary>
+        /// <param name="registerRequest">Register data object</param>
```
jonasanker, line 38: Register data request?

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -94,31 +94,29 @@ public string Login(string email, string password, string version)
...
-                Name = EscapeName(registerDto.Name),
-                Email = registerDto.Email,
+                Name = EscapeName(name),
+                Email = email,
                 Password = hashedPassword,
                 Salt = salt,
-                Programme = programme,
+                Programme = chosenProgramme,
                 UserGroup = UserGroup.Customer
             };
```
jonasanker, line 122: Line 124, what is the case for when it makes sense to wait a minute?

jonasanker: Maybe it makes sense to call just call SaveChangesAsync without if?

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Services/AccountService.cs
Hjaltesorgenfrei: Yeah, its for backwards compatibility, so the call in the other controller does not need to add it.

coffeecard/CoffeeCard.Library/Services/IAccountService.cs
```diff
@@ -9,7 +9,8 @@ namespace CoffeeCard.Library.Services
     public interface IAccountService
     {
         User GetAccountByClaims(IEnumerable<Claim> claims);
-        Task<User> RegisterAccountAsync(RegisterDto registerDto);
+        // Programme 1 is SWU
```
jonasanker, line 12: Use the code documentation style for methods comments, e.g.

```csharp
/// <summary>
/// This property always returns a value &lt; 1.
/// </summary>
/// <param name="programme">Programme Id of ITU Study programme. Default value is 1 (for SWU) for backwards compability</param>
...
```

jonasanker (changes requested):

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/RegisterDTO.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    /// <example>
+    /// {
+    ///     "name": "John Doe",
+    ///     "email": "john@doe.com",
+    ///     "password": "[no example provided]",
+    ///     "programme": 1
+    /// }
+    /// </example>
+    public class RegisterAccountRequest
```
jonasanker, line 16: The filename and class name differs. I prefer the name of the class name

jonasanker (changes requested):
One small comment left

jonasanker (approved):

PR #108, opened 2022-11-03
Better verification of jwt token. Better error handling on web pages

TTA777:


jonasanker:
I'll review this in the weekend as I want to have a better look

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,24 +44,68 @@ public JwtSecurityToken ReadToken(string token)
...
+                {
+                    Log.Information("Received invalid token");
+                    return false;
+                }
+
+                return true;
+            }
+            catch (ArgumentException e)
+            {
+                Log.Error($"Unable to read token. Exception thrown = {e}");
```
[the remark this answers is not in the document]
TTA777: Not a log statement on 68, but changed the one on 113

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,28 +44,73 @@ public JwtSecurityToken ReadToken(string token)
...
+                        ValidateLifetime = true,
+                        ClockSkew = TimeSpan.Zero //the default for this setting is 5 minutes
+                    };
+
+                    securityTokenHandler.ValidateToken(tokenString, validationParameters, out _); // Throws exception if token is invalid
+                }
+                catch (Exception e) when (e is ArgumentException ||
+                                          e is SecurityTokenException)
+                {
+                    Log.Information("Received invalid token");
```
jonasanker, line 79: For what reason would an ArgumentException be thrown and does it make sense to log the same message in this case? If it is unexpected to receive an ArgumentException, I think it's important to log the stack trace

coffeecard/CoffeeCard.WebApi/Pages/VerifyEmail.cshtml.cs
```diff
@@ -16,11 +17,11 @@ public VerifyEmail(IAccountService accountService)
...
-        public IActionResult OnGet()
+        public async Task<IActionResult> OnGet()
         {
             try
             {
-                var emailVerified = _accountService.VerifyRegistration(Token);
+                var emailVerified = await _accountService.VerifyRegistration(Token);
                 if (emailVerified)
                 {
                     TempData["resultHeader"] = "Success";
```
jonasanker, line 27: Comment for the `catch` block on line 32. Why do we catch the Exception supertype here? Would it not make more sense to use an `else`

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
```diff
@@ -39,12 +47,15 @@ public async Task<IActionResult> OnPostAsync()
...
-                TempData["resultHeader"] = "Success";
-                TempData["result"] = "Your password has now been reset";
+                if (await _accountService.RecoverUserAsync(Token, PinCode.NewPinCode))
+                {
+                    TempData["resultHeader"] = "Success";
+                    TempData["result"] = "Your password has now been reset";
+                }
             }
-            else
+            catch (Exception e)
```
jonasanker, line 58: Why the try/catch here, what is the scenario for this?

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
```diff
@@ -25,9 +26,16 @@ public Recover(IAccountService accountService, ITokenService tokenService)
...
+            {
+                IsTokenValid = await _tokenService.ValidateTokenIsUnusedAsync(Token);
 
-            if (IsTokenValid) return Page();
+                if (IsTokenValid) return Page();
+            }
+            catch (Exception e)
+            {
+                // ignored
+            }
```
jonasanker, line 38: This is not an optimal pattern. Why do we catch the Exception? If at all, we must at least log the exception else we would never know the exception was thrown

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,28 +44,73 @@ public JwtSecurityToken ReadToken(string token)
...
+                return false;
+            }
+        }
+        
+        /// <summary>
+        /// Receives the serialized version of a JWT token as a string.
+        /// Checks if the JWT token is valid (Based on its lifetime and if it has been used before)
+        /// </summary>
+        /// <param name="tokenString"></param>
+        /// <returns></returns>
```
jonasanker, line 97: Missing entry in returns documentation block

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -265,12 +265,12 @@ public async Task RequestAnonymization(User user)
             
             await _emailService.SendVerificationEmailForDeleteAccount(user, verificationToken);
         }
-        
-        public async void AnonymizeAccount(string token)
+
+        public async Task AnonymizeAccountAsync(string token)
         {
             Log.Information($"Trying to verify deletion with token: {token}");
```
jonasanker, line 271: No need for string interpolation when using SeriLog

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -289,20 +289,25 @@ private async Task AnonymizeUser(User user)
...
             user.UserState = UserState.Deleted;
-            await _context.SaveChangesAsync();            
+            await _context.SaveChangesAsync();
         }
-        
-        private User VerifyTokenClaimAndUser(string token)
+
+        private async Task<User> VerifyTokenClaimAndUserAsync(string token)
         {
+            if (await _tokenService.ValidateTokenAsync(token))
```
jonasanker, line 297: Should it not be the negated case?

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -289,20 +289,25 @@ private async Task AnonymizeUser(User user)
...
-
+                throw new ApiException("The token is invalid!", StatusCodes.Status401Unauthorized);
+            
             var emailClaim = jwtToken.Claims.FirstOrDefault(x => x.Type == ClaimTypes.Email);
-            if (emailClaim == null) throw new ApiException("The token is invalid!", 400);
+            if (emailClaim == null) throw new ApiException("The token is invalid!", StatusCodes.Status401Unauthorized);
 
             var user = _context.Users.FirstOrDefault(x => x.Email == emailClaim.Value);
-            if (user == null) throw new ApiException("The token is invalid!", 400);
+            if (user == null) throw new ApiException("The token is invalid!", StatusCodes.Status401Unauthorized);
```
jonasanker, line 310: Since the ApiException message is the same for all three, but different, cases, we should at least log an info or debug with the reason for failing else we would not be able to tell why the token failed

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,28 +44,73 @@ public JwtSecurityToken ReadToken(string token)
...
 
+        
         /// <summary>
-        ///     Receives the serialized version of a JWT token as a string. Checks if the JWT token is valid (Based on its lifetime
-        ///     and if it has been used before)
+        /// Receives the serialized version of a JWT token as a string.
+        /// Checks if the JWT token is valid (Based on its lifetime)
         /// </summary>
         /// <param name="tokenString"></param>
         /// <returns></returns>
```
jonasanker, line 53: Missing returns code documentation

coffeecard/CoffeeCard.Library/Services/AccountService.cs
jonasanker: I guess you could have a test case for this one 😉

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Services/TokenService.cs
TTA777: ArgumentException will be thrown if the token given for validation is malformed in some way, e.g., can't be read, is null, or if the validation parameters are null. I think it is safe to handle this in the same manner as an invalid token. 

I missed there being a second try catch surrounding this bit when I made it though, where it does catch ArgumentException, so we could also remove it from this catch and rely on that instead. Or remove the second one entirely, either way something needs to be changed.

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
TTA777: Same reason as above. The is catch for any unhandled Exceptions from the service. Otherwise, if an exception does occur, it will in the best case be propagated to the user as a plain json website, and worst case bring down the entire site (Happened once in development).

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
TTA777: I'll add a log statement. We catch for the same reason as the two above

coffeecard/CoffeeCard.Library/Services/AccountService.cs
TTA777: Added some tests, turns out one of the other cases was also wrong

TTA777:
I think all of your previous comments have been addressed now, but I'm leaving them open so you can go through and check

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -265,13 +274,22 @@ public async Task RequestAnonymization(User user)
...
+            User user;
+            try
+            {
+                user = await VerifyTokenClaimAndUserAsync(token);
+            }
+            catch (ApiException e)
+            {
+                // No log, since the ApiException should have been logged on throwing
+                return;
+            }
```
jonasanker, line 291: Why do we supress the exception? I assume we want to return the 401 to the user?

jonasanker: Unless we want to suppress he exception, I suggest to remove the catch block since we are not changing logic if exception is thrown

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -138,14 +138,23 @@ public async Task<User> RegisterAccountAsync(RegisterDto registerDto)
...
+            User user;
+            try
+            {
+                user = await VerifyTokenClaimAndUserAsync(token);
+            }
+            catch (ApiException e)
+            {
+                // No log, since the ApiException should have been logged on throwing
+                return false;
+            }
```
jonasanker, line 154: Same question as earlier, why are we suppressing the exception?

coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -289,21 +307,38 @@ private async Task AnonymizeUser(User user)
             user.DateUpdated = DateTime.Now;
             user.PrivacyActivated = true;
             user.UserState = UserState.Deleted;
-            await _context.SaveChangesAsync();            
+            await _context.SaveChangesAsync();
         }
-        
-        private User VerifyTokenClaimAndUser(string token)
+
+        private async Task<User> VerifyTokenClaimAndUserAsync(string token)
```
jonasanker, line 313: Not on you, but doesn't it seem a bit weird that we read claims in AccountService and not in the TokenService?

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,28 +44,65 @@ public JwtSecurityToken ReadToken(string token)
...
+                    ValidateLifetime = true,
+                    ClockSkew = TimeSpan.Zero //the default for this setting is 5 minutes
+                };
+
+                securityTokenHandler.ValidateToken(tokenString, validationParameters, out _); // Throws exception if token is invalid
+            }
+            catch (Exception e) when (e is ArgumentException ||
+                                      e is SecurityTokenException)
+            {
+                Log.Information("Received invalid token");
```
jonasanker, line 77: Do you think it would make sense to log more, e.g. the stack trace, and the token (maybe in a debug log statement) just to have some logging and debug hints if this start failing?

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
```diff
@@ -25,9 +27,16 @@ public Recover(IAccountService accountService, ITokenService tokenService)
...
+            {
+                IsTokenValid = await _tokenService.ValidateTokenIsUnusedAsync(Token);
 
-            if (IsTokenValid) return Page();
+                if (IsTokenValid) return Page();
+            }
+            catch (Exception e)
+            {
+                Log.Error("An unknown exception {ex} occured while validating token {token} for account recovery", e, Token);
+            }
```
jonasanker, line 39: I think we must decide what pattern we go for when it comes to handling the invalid token and exception handling. 
We do it differently on different pages. Here we do if (no else), then a catch with logging and treat the invalid token and treat any exception the same when communicating to the user.
But e.g. in `VerifyEmail.cshtml.cs` we do have the else case, we catch the exception but do not log, and treat the invalid email error message and caught the exception error message uniquely.

On top of my head, I suggest that we apply a pattern closer to the last for **all** pages
1. Both if and else case
2. Catch Exception, log the exception
3. For failures, i.e. the _else_ case and the exception, the error message is set uniquely and describes something meaningful for the user

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,28 +44,65 @@ public JwtSecurityToken ReadToken(string token)
...
-        ///     Receives the serialized version of a JWT token as a string. Checks if the JWT token is valid (Based on its lifetime
-        ///     and if it has been used before)
+        /// Receives the serialized version of a JWT token as a string.
+        /// Checks if the JWT token is valid (Based on its lifetime)
         /// </summary>
         /// <param name="tokenString"></param>
-        /// <returns></returns>
-        public async Task<bool> ValidateToken(string tokenString)
+        /// <returns>bool</returns>
+        public async Task<bool> ValidateTokenAsync(string tokenString)
```
jonasanker, line 54: No asynchronous code in this method. Async can be removed

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,28 +44,65 @@ public JwtSecurityToken ReadToken(string token)
...
-                if (user.Tokens.Contains(new Token(tokenString))) tokenNotUsed = true;
+                if (user.Tokens.Contains(new Token(tokenString))) tokenIsUnused = true; // Tokens are removed from the user on account recovery
 
-                return token.ValidTo > DateTime.UtcNow && tokenNotUsed;
+                return await ValidateTokenAsync(tokenString) && tokenIsUnused;
             }
             catch (ArgumentException e)
             {
-                Log.Error($"Unable to read token. Exception thrown = {e}");
+                Log.Error("Unable to read token. Exception thrown = {}", e);
```
jonasanker, line 105: Apparently the syntax for logging exceptions is
```csharp
Log.Error(e, "Unable to read token. Exception thrown",);
```

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Services/TokenService.cs
TTA777: The token itself could make sense to log in a debug statement, but I find it very unlikely that we will be looking at it. I find it unlikely that this check will start failing, since we are making use of the same validation parameters here, that we use on the API as a whole.

coffeecard/CoffeeCard.Library/Services/TokenService.cs
TTA777: Removed async

coffeecard/CoffeeCard.Library/Services/AccountService.cs
TTA777: Hmm, the reasoning for it was this method is only called from the webpages, so to limit the amount of exceptions they receive. This was probably the wrong approach, and the better way would be to catch the ApiException separately on the webpages and handle it by showing an appropriate error message

coffeecard/CoffeeCard.Library/Services/AccountService.cs
TTA777: I agree, I suppose the original reasoning would have been it returns a user object, hence its placement in accountService. I suppose we could move it to the tokenSerive as a getEmailFromVerificationToken, or something like that. Then we could still let the accountService handle the retrieval of the suer object.

jonasanker: You can decide if you want to do anything about this or not :)

coffeecard/CoffeeCard.Library/Services/AccountService.cs
jonasanker: I agree with what you suggest as your solution. Ignore it here, catch at web pages 😉

coffeecard/CoffeeCard.Library/Services/AccountService.cs
TTA777: Moved to TokenService

coffeecard/CoffeeCard.WebApi/Pages/Recover.cshtml.cs
TTA777: This got refactored a bit. Now each page gives a Func to a PageUtil that then executes the func, and handles any exceptions that my occur from running the code. Let me know what you think of this approach

coffeecard/CoffeeCard.Library/Services/TokenService.cs
```diff
@@ -44,30 +47,92 @@ public JwtSecurityToken ReadToken(string token)
...
             }
             catch (ArgumentException e)
             {
-                Log.Error($"Unable to read token. Exception thrown = {e}");
+                Log.Error(e, "Unable to read token. Exception thrown");
                 return false;
             }
         }
+        
+        public string ValidateTokenAndGetEmail(string token)
```
jonasanker, line 113: Consider renaming to `ValidateVerificationTokenAndReadEmail` since method only applies for verification tokens

coffeecard/CoffeeCard.WebApi/Helpers/PageUtils.cs
```diff
@@ -0,0 +1,50 @@
...
+            try
+            {
+                return await Task.Run(func);
+            }
+            catch (ApiException ex)
+            {
+                // Not logging this, since we should have logged the cause when we created it.
+                if (ex.StatusCode == StatusCodes.Status401Unauthorized)
+                {
+                    setMessage("Error", "Looks like the link you used has expired or already been used. Request a new link and try again", model);
```
jonasanker, line 28: It is not clear to me in what case we throw a 401 for an already-used token?

jonasanker (dismissed):
Approved with a few minor comments which you can consider if you want to include

coffeecard/CoffeeCard.WebApi/Helpers/PageUtils.cs
TTA777: Currently, we do not, though we might in the future if we make the email confirmation and delete account methods, check if the token has been used

jonasanker: Mjah :(

jonasanker (approved):

Issue #121, opened 2022-12-28
Include ProductName in SimplePurchaseResponse


jonasanker:
Include ProductName property in the SimplePurchaseResponse class

TTA777:
If this is for the receipts, then the amount of tickets provided by the product, and the price of it will also be required.

PR #122, opened 2022-12-28
Update Purchase API v2

### LINKED ISSUE #120: GetPurchase endpoint fails when payment was not made using mobilepay
GetPurchase calls MobilePayPaymentsService.GetPayment no matter the payment type used for the purchase. This results in the request failing if another payment service was used, e.g. a free product

jonasanker:
Closes #120 , #121

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/PurchaseService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/PurchaseService.cs
@@ -131,8 +131,9 @@ namespace CoffeeCard.Library.Services.v2
                 OrderId = paymentDetails.OrderId,
                 TransactionId = transactionId,
                 PurchasedBy = user,
-                Status = purchaseStatus
-                // FIXME State management, PaymentType
+                Status = purchaseStatus,
+                PaymentType = purchaseRequest.PaymentType
+                // FIXME State management
             };
             
             return (purchase, paymentDetails);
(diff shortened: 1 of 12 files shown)


coffeecard/CoffeeCard.Library/Services/v2/PurchaseService.cs
```diff
@@ -152,7 +153,26 @@ public async Task<SinglePurchaseResponse> GetPurchase(int purchaseId, User user)
                     $"No purchase was found by Purchase Id: {purchaseId} and User Id: {user.Id}");
             }
 
-            var paymentDetails = await _mobilePayPaymentsService.GetPayment(Guid.Parse(purchase.TransactionId));
+            PaymentDetails paymentDetails;
+            if (purchase.PaymentType == null)
```
TTA777, line 157: From my understanding, this assumes that all purchases done prior to this pull request were done using mobilePay. This however is not correct, due to us having issued vouchers as part of the datamodel.  As such, I believe line 159 will result in unintended behaviour.

TTA777 (changes requested):
I believe there is one severe issue

TTA777:
@jonasanker This has been stale for quite a while. Do we want it fixed, and merged, or do we want a rework of the er model as a "proper" fix instead?

jonasanker:
@TTA777 Yea, I think it was lost when the focus changed to Azure. Can we maybe discuss next time or online how the new data model might look like so we can get a sense of the amount of changes?

jonasanker:
Awaits completion of #130

marfavi:
Branch is now up to date. What now❓ @jonasanker @TTA777

coffeecard/CoffeeCard.Library/Services/v2/PurchaseService.cs
marfavi: addressed in latest commit

PR #134, opened 2023-03-02
Implemented get and update endpoints for AccountController v2 

TTA777:
- Refactored the service layer, so account controller v2 no longer relies on the v1 account service. 
- Added get and update to account controller v2. The new functionality is it now includes all information about a user's programme, not only the ID of it.
- Removed methods from account service v1 that supported functionality in account controller v2
-  Added tests for all methods in AccountService v2

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/IAccountService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/IAccountService.cs
@@ -15,11 +15,7 @@ namespace CoffeeCard.Library.Services.v2
         /// </summary>
         /// <param name="programme">Programme Id of ITU Study programme.
         Task<UserResponse> RegisterAccountAsync(string name, string email, string password, int programme);
-        Task<bool> VerifyRegistration(string token);
         User UpdateAccount(IEnumerable<Claim> claims, UpdateUserRequest userDto);
-        void UpdateExperience(int userId, int exp);
-        Task ForgotPasswordAsync(string email);
-        Task<bool> RecoverUserAsync(string token, string newPassword);
         Task RequestAnonymization(User user);
         Task AnonymizeAccountAsync(string token);
         Task<bool> EmailExists(string email);
--- a/coffeecard/CoffeeCard.Library/Services/v2/IAccountService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/IAccountService.cs
@@ -1,0 +1,27 @@
+using System.Collections.Generic;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using CoffeeCard.Models.DataTransferObjects.v2.User;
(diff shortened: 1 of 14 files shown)


coffeecard/CoffeeCard.Library/Services/v2/IAccountService.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        /// <summary>
+        /// Register an account. Sends out an verification email to provided email
+        /// </summary>
+        /// <param name="programme">Programme Id of ITU Study programme.
+        Task<UserResponse> RegisterAccountAsync(string name, string email, string password, int programme);
+        User UpdateAccount(IEnumerable<Claim> claims, UpdateUserRequest userDto);
+        Task RequestAnonymization(User user);
+        Task AnonymizeAccountAsync(string token);
+        Task<bool> EmailExists(string email);
```
fredpetersen, line 21: Vil man ikke gerne have xml-doc på resten af dette også? :)

coffeecard/CoffeeCard.Library/Services/v2/ILeaderboardService.cs
```diff
@@ -22,5 +22,8 @@ public interface ILeaderboardService
         /// <param name="preset">Preset to filter</param>
         /// <returns>Leaderboard entry for user</returns>
         Task<LeaderboardEntry> GetLeaderboardEntry(User user, LeaderboardPreset preset);
+
+        //Todo add desciption
+        Task<(int Total, int Semester, int Month)> GetLeaderboardPlacement(User user);
```
fredpetersen, line 27: beskrivelse her også

fredpetersen (commented):
få kommentarer

jonasanker:
> * Currently, the account service v2 relies on other services, some of which are v1. Should this also be moved to v2 as part of this PR?

Let's be pragmatic here and only focus on the AccountServices v2

> * We have been somewhat inconsistent in what types are returned from, and given to, the service layer, i.e., we are using both entities, DTOs and simple types for this. What do we want to do going forward?

While this is not documented, I have followed these principles for v2 services

1. At the latest on the Controller, we should transform the Entity. I don't see such a big problem with services taking Entity classes as input as long as we do not expose the Entity to externals. What we would like to safe guard is the freedom to change data model at any time without coordination with externals.
2. Always use classes as request/response types. This gives us the freedom to extend the class without introducing breaking changes.
3. We do not name it DTO, we use a meaningful name for its purpose, e.g. AccountCreationRequest, AccountCreationResponse etc

> * Should the test be written as part of this pr? There were none for the code that was moved from v1, to v2. Should we add some there as well?

I think it's important that we improve the code base quality in v2. This was part of my motivation to start the rewrite. With this intention, we should write tests to Account Service 2

fremartini (approved):

PR #151, opened 2023-04-25
Analog Core goes to Azure

jonasanker:
- Use Azure App Services, Azure SQL, Key Vault and Application Insights
- Migrate to .NET 6


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
+name: Deploy
+
+on:
+  push:
+    branches: [ develop, production, feature/azure-deploy ]
```
jonasanker, line 5: Should be removed before merging

PR #168, opened 2023-05-18
Minor tweaks to Pull Request

jonasanker:
- Added index targeting to GetSortedStatistics query
- Improve documentation on the fly

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
@@ -23,10 +23,10 @@ namespace CoffeeCard.Library.Services.v2
 
         public async Task<IEnumerable<LeaderboardEntry>> GetTopLeaderboardEntries(LeaderboardPreset preset, int top)
         {
-            var sortedStatistics = GetSortedStatistics(preset)
+            var sortedStatistics = await GetSortedStatistics(preset)
                 .Take(top)
                 .Include(s => s.User)
-                .AsEnumerable();
+                .ToListAsync();
 
             return sortedStatistics.Select((s, index) => new LeaderboardEntry
             {
@@ -75,7 +75,8 @@ namespace CoffeeCard.Library.Services.v2
             return leaderBoardPlacement;
         }
 
-        private IOrderedQueryable<Statistic> GetSortedStatistics(LeaderboardPreset preset){
+        private IOrderedQueryable<Statistic> GetSortedStatistics(LeaderboardPreset preset)
+        {
             var statPreset = preset.ToStatisticPreset();
 
             return _context.Statistics
--- a/coffeecard/CoffeeCard.Models/Entities/Statistic.cs
+++ b/coffeecard/CoffeeCard.Models/Entities/Statistic.cs
@@ -1,31 +1,73 @@
 ﻿using System;
 using System.ComponentModel.DataAnnotations.Schema;
+using Microsoft.EntityFrameworkCore;
 
 namespace CoffeeCard.Models.Entities
 {
-    /**
-    * This class will be used to optimize the queries to get statistics.
-    * The preset defines the kind of statistic entity and each preset has a corresponding `SwipeCount`, `SwipeRank` and `ExpiryDate`.
-    * This means that each user should have exactly 3 entries in the database after 1 or more swipes.
-    * Each preset should be updated accordingly to the current date and the ExpiryDate. If the ExpiryDate of a entry is exceeded, then the `SwipeCount`, `SwipeRank` and `ExpiryDate` should be reset.
-    **/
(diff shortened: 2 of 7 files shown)


coffeecard/CoffeeCard.Models/Entities/Statistic.cs
```diff
@@ -1,31 +1,73 @@
...
+        /// <summary>
+        /// All Statistics
+        /// </summary>
+        Total
+    }
 
+    /// <summary>
+    /// Statistic represent a user's swipe of tickets. A Statistic entry is valid within a given preset period unless it is the total statistics.
+    /// </summary>
+    [Index(nameof(Preset), nameof(ExpiryDate), nameof(SwipeCount), nameof(LastSwipe))]
```
TTA777, line 34: My main concern with this index, would be whether it is too expensive to maintain for the database, compared to the increase in speed it would provide. 

That is, every time a person makes a swipe we would end up having to rebuild the index. Have you done any testing to see how this performs?

I considered adding one to the preset myself, since this should not change, so there would be less overhead in maintaining it. We could also expand it to include the expiry date, since this should change rarely as well.

coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
```diff
@@ -23,10 +23,10 @@ public LeaderboardService(CoffeeCardContext context, IDateTimeProvider dateTimeP
 
         public async Task<IEnumerable<LeaderboardEntry>> GetTopLeaderboardEntries(LeaderboardPreset preset, int top)
         {
-            var sortedStatistics = GetSortedStatistics(preset)
+            var sortedStatistics = await GetSortedStatistics(preset)
                 .Take(top)
                 .Include(s => s.User)
-                .AsEnumerable();
+                .ToListAsync();
```
TTA777, line 29: I was seeing slightly worse performance on my setup using this instead of enumerable. Presumably because it has to load the entire thing into memory instead. Did you see differently? 

If it is because we want it to be async, I think we should look into the AsyncEnumerable and the nuget package that adds support for linq statements on it.

TTA777 (changes requested):

coffeecard/CoffeeCard.Models/Entities/Statistic.cs
jonasanker: I don't think that it will be expensive compared to the size of the table. I imagined this index would be useful for the Get Sorted Statistics.

I can't recall the documentation entirely but don't think the entire Index is being rebuilt at every swipe, only the relevant entries. Has this been a sorted index then likely it could be that it would be completely rebuilt.

coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
jonasanker: Honestly, it was to asynchronousity.  

After researching a bit, I see a few recommendations to force query evaluation by using `ToListAsync()` at the earliest point where no more filters on the database query are made, see here [[1]](https://anduin.aiursoft.com/post/2020/1/26/tips-to-get-better-performance-for-entity-framework-core) and somewhat related [[2]](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/best-practices?view=aspnetcore-6.0#return-ienumerablet-or-iasyncenumerablet)

TTA777: Did you do any kind of testing regarding the performance? Again, I was seeing better performance on my system with it returning an enumerable, although that is also in a situation where there is only one client making requests, so it might very well be different in production.

coffeecard/CoffeeCard.Models/Entities/Statistic.cs
```diff
@@ -1,31 +1,73 @@
...
     {
+        /// <summary>
+        /// Statistic Id
+        /// </summary>
         public int Id { get; set; }
         
+        /// <summary>
+        /// Preset. A uesr is expected to have one entry of each Preset type
+        /// </summary>
         public StatisticPreset Preset { get; set; }
```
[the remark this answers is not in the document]
TTA777: So do we want to do that as part of this request, since we are changing the model anyway, or do it at some other time?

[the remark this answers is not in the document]
jonasanker: Let's do it later, so you can get your PR out

coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
jonasanker: Per the [2], we shouldn't return an IEnumerable from the controller level, so either here or on the Controller we should apply the ToList()

TTA777 (approved):

PR #178, opened 2023-06-06
style: format files

marfavi:
run `dotnet format` in the `coffeecard` dictionary


coffeecard/CoffeeCard.WebApi/appsettings copy.bicep
```diff
@@ -0,0 +1,177 @@
+// {
```
TTA777, line 1: Where did this file come from?

TTA777 (changes requested):
All good, except one mystery file

jonasanker:
Great initiative! Maybe it could be useful to add a pipeline CI step where we verify that all files are formatted. Similar to what we are doing in the coffeecard_app repository.

The command
```
dotnet format --verify-no-change
```

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format#options

coffeecard/CoffeeCard.WebApi/appsettings copy.bicep
marfavi: probably leftover file from the last time i was in this repo. removing

marfavi:
> Great initiative! Maybe it could be useful to add a pipeline CI step where we verify that all files are formatted. Similar to what we are doing in the coffeecard_app repository.
> 
> The command
> 
> ```
> dotnet format --verify-no-change
> ```
> 
> https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format#options

#180

PR #180, opened 2023-06-07
ci: Check file formatting in pipeline

marfavi:


.github/workflows/core-build.yml
```diff
@@ -7,35 +7,51 @@ on:
...
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+      - name: Restore dependencies
+        run: dotnet restore coffeecard/
+      - name: Build CoffeeCard solution
+        run: dotnet build coffeecard/ --no-restore /p:ContinuousIntegrationBuild=true
+
+  test:
```
jonasanker, line 31: We end up building the solution **twice** by separating the build and test into two separate jobs. This will extend the pipeline execution time.

I suggest that we keep them in the same job

jonasanker (changes requested):

marfavi:
I suggest #178 is merged first so we can see this workflow complete without errors

jonasanker (approved):
LGTM

PR #183, opened 2023-06-08
refactor: Reduce amount of warnings from 180 to 17

marfavi:
- Added doc comments to all public fields (phew!)

- Added nonempty constructors to entities where a field could be null after creation (like strings).
  > Some of these constuctors throw an error if app attempted to instantiate with empty values.
  > Entities still have 0-argument constructors to work with DbSet.

- Some entities now have simplified constructors.
  > For example, the new Purchase constructor sets 8 fields using 3 parameters.
  > The new Ticket constructor sets 5 fields with 1 parameter.

- Fixed tests as a result of breaking them.
  > Enable `nullable` in test project
  > Most should be more robust by avoiding magic strings
  > Test object getters, like a `testUser` getter, reduces code duplication
  > Redundant fields in test objects have been removed while ensuring the tests still pass.

- Removed `LoginAttempt` entity as stated in a TODO comment
- Removed pragma warning ignores
- Removed unused usings


jonasanker:
Instead of using constructors. we want to use the new required property introduced in C# 11. This will be available in #191. Hereafter we suggest to use this instead of constructors, this we pause this until the completion #191

TTA777:
Closed, since #310 was merged in favour of this

PR #185, opened 2023-06-11
Fix RedeemVoucher check on voucher usage

### LINKED ISSUE #184: A voucher can be redeemed multiple times
Possible related to `PurchaseServices.cs l276` where there is no Include statement

jonasanker:
Closes #184


marfavi (changes requested):
Write a test that would've caught the faulty behaviour. Verify that your test fails on branch develop and succeeds on this branch

TTA777 (approved):
lgtm, file issue on the testing

PR #190, opened 2023-07-15
Cleanup of purchase data

### LINKED ISSUE #130: Redesign data model for Purchases, Vouchers and Free Products
Hi @fredpetersen
@marfavi and I discussed this. Omid's argumentation is that not all Purchases have a Payment and hence strictly speaking not a purchase. This is for instance the case with vouchers where a user is not _purchasing_ anything but he's using a voucher which grants him Tickets. The point is that our data model is not reflecting this and it would have been best if it did. However this will require both a larger redesign and secondly a data migration. So create an issue from this and you can go ahead and merge this Pull Request. Yay 🙌🏻

_Originally posted by @jonasanker in https://github.com/AnalogIO/analog-core/pull/119#discussion_r1107341767_

TTA777:
Closes #130 

The migrations have been tested on a both a local copy of current prd and dev data, and appear to work as intended

Purchase type should also be set in all places where appropriate in the code


coffeecard/CoffeeCard.Library/Migrations/20230711144034_AddPurchaseTypes.cs
```diff
@@ -0,0 +1,97 @@
...
+                table: "Vouchers",
+                type: "int",
+                nullable: true);
+
+            migrationBuilder.AddColumn<string>(
+                name: "Type",
+                schema: "dbo",
+                table: "Purchases",
+                type: "nvarchar(max)",
+                nullable: false,
```
jonasanker, line 23: We probably need to do this in two steps; one where Type is nullable and allows us to populate the column. A second where we set it to non nullable

coffeecard/CoffeeCard.Library/Migrations/20230711161456_PurchaseDataCleanup.cs
```diff
@@ -0,0 +1,33 @@
...
+            migrationBuilder.Sql("UPDATE dbo.Purchases SET Type = 'PointOfSale' Where OrderId like 'Analog'");
+
+            // Update vouchers with references to the purchase issued by its redemption
+            migrationBuilder.Sql("update Vouchers set PurchaseId = Purchases.Id from Purchases where Purchases.OrderId = Vouchers.Code");
+
+            // Create new vouchers for purchases where the same voucher have been used to issue multiple purchase
+            migrationBuilder.Sql("insert into Vouchers (Code, DateCreated, DateUsed, Description, Requester, Product_Id, User_Id, PurchaseId) select Purchases.OrderId + '-' + cast(Purchases.Id as varchar), Purchases.DateCreated, Purchases.DateCreated, 'Creation of extra vouchers for purchases without references', 'AnalogIO', Purchases.ProductId, Purchases.PurchasedBy_Id, Id from Purchases where Type = 'Voucher' and Id not in (select PurchaseId from Vouchers where PurchaseId is not null)");
+            // Insert data to the new Pos table
+            migrationBuilder.Sql("insert into PosPurchases (PurchaseId, BastiaInitials) select Id, TransactionId from Purchases where OrderId like 'Analog'");
```
jonasanker, line 25: Spelling error -> BaristaInitials

coffeecard/CoffeeCard.Library/Migrations/20230714211716_PurchaseOrderIdUniqueAndExternalTransactionId.cs
```diff
@@ -0,0 +1,65 @@
...
+#nullable disable
+
+namespace CoffeeCard.Library.Migrations
+{
+    public partial class PurchaseOrderIdUniqueAndExternalTransactionId : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Make transactionId null, where not MobilePayV1, or MobilepayV2, and generate unique Guid orderIds
+            migrationBuilder.Sql("update Purchases set OrderId = NEWID(), TransactionId=null where Type != 'MobilePayV1' and Type != 'MobilePayV2' or OrderId = 'OLD PURCHASES'");
```
jonasanker, line 12: Where do we set this?

coffeecard/CoffeeCard.Models/Entities/PosPurchase.cs
```diff
@@ -0,0 +1,8 @@
+namespace CoffeeCard.Models.Entities
+{
+	public class PosPurhase{
+		public int Id { get; set; }
+		public Purchase Purchase { get; set; }
+		public string BastiaInitials {get; set; }
```
jonasanker, line 6: Spelling -> BaristaInitials

coffeecard/CoffeeCard.Models/Entities/PosPurchase.cs
```diff
@@ -0,0 +1,8 @@
+namespace CoffeeCard.Models.Entities
+{
+	public class PosPurhase{
```
jonasanker, line 3: Run a formatter on this and add documentation

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Migrations/20230714211716_PurchaseOrderIdUniqueAndExternalTransactionId.cs
TTA777: Not sure what you mean? Set is the syntax for updating a specific column in sql, and NEWID() is a build in function of sql server, so we are essentially updating all purchases with a new guid as the orderId where it wasn't already.

coffeecard/CoffeeCard.Library/Migrations/20230711144034_AddPurchaseTypes.cs
```diff
@@ -0,0 +1,97 @@
...
+                defaultValue: "");
```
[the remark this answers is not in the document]
TTA777: Not sure if I see much difference in being able to check against the empty string, and null, though I can change it if you really want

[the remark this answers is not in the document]
jonasanker: The difference is that when you have a default value then the database will use this value regardless if it's intended or not. When you have no default value, the insert call will fail should you forget to add a value for the column. We want this

coffeecard/CoffeeCard.Library/Migrations/20230714211716_PurchaseOrderIdUniqueAndExternalTransactionId.cs
jonasanker: Sorry. I meant the 'Old purchases'. Do we have entries with this name?

TTA777: I think there were like 2 or 3 of them in the production dataset. I'm guessing they are the result of someone moving some data manually at one point or another. Not something we set from the code

coffeecard/CoffeeCard.Library/Migrations/20230711144034_AddPurchaseTypes.cs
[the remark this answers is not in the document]
TTA777: Gotcha, so not a concern in regard to this clean-up, but more in going forward. I'll have a look at how to change this without too much pain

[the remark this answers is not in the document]
jonasanker: It's related to this clean up in that sense that I thin it should not have a default value at all :wink:

coffeecard/CoffeeCard.Models/Entities/PosPurchase.cs
```diff
@@ -0,0 +1,13 @@
+namespace CoffeeCard.Models.Entities
+{
+    /// Represents a purchase issued by a barista at the point of sale in the cafe
+    public class PosPurhase
+    {
+        /// Primary key
+        public int Id { get; set; }
```
jonasanker, line 7: Does the Id make sense, will it not always be the PurchaseId which is the primary id?

coffeecard/CoffeeCard.Models/Entities/PosPurchase.cs
```diff
@@ -0,0 +1,13 @@
...
+        /// The purchase that was issued by the barista
+        public Purchase Purchase { get; set; }
```
jonasanker, line 9: I think we would need a property PurchaseId which we can query on to avoid having to eagerly load the Purchase property if we need to work with it

coffeecard/CoffeeCard.Models/Entities/Voucher.cs
```diff
@@ -29,5 +29,7 @@ public class Voucher
         public int? UserId { get; set; }
 
         public User? User { get; set; }
+
+        public Purchase? Purchase { get; set; }
```
jonasanker, line 33: Would we need the property PurchaseId for same reason as for PosPurchases?

jonasanker (changes requested):

jonasanker (approved):

PR #204, opened 2023-09-14
fix: Correct where clause in resend verification

duckth:
According to [this article from the EF Core docs](https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity#translation-of-built-in-net-string-operations), using the `StringComparison` enum in `Equals` does not correctly translate to SQL, and as such EF Core will throw an exception. 

This pull request changes the `Where` clause to simply perform `.ToLower()` on each string, to ensure case-insensitivity


marfavi (approved):
OK. Please verify that the solution works after merge

PR #211, opened 2023-09-23
Use unique OrderId for Vouchers

jonasanker:


coffeecard/CoffeeCard.Library/Services/PurchaseService.cs
```diff
@@ -22,7 +23,7 @@ public PurchaseService(CoffeeCardContext context)
             _context = context;
         }
 
-        public Purchase RedeemVoucher(string voucherCode, IEnumerable<Claim> claims)
+        public async Task<Purchase> RedeemVoucher(string voucherCode, IEnumerable<Claim> claims)
```
TTA777, line 26: Should be named Async

coffeecard/CoffeeCard.Library/Services/PurchaseService.cs
```diff
@@ -63,6 +64,20 @@ public Purchase RedeemVoucher(string voucherCode, IEnumerable<Claim> claims)
             return purchase;
         }
 
+        private async Task<Guid> GenerateUniqueOrderId()
```
TTA777, line 67: Normally I would say refactor to avoid code duplication here, but do we want to care considering it will likely be deprecated soon? I'm guessing that's why you didn't?

TTA777 (changes requested):

coffeecard/CoffeeCard.Library/Services/PurchaseService.cs
jonasanker: The old purchase service will be deprecated as soon as we app is not using that Endpoint

coffeecard/CoffeeCard.Library/Services/PurchaseService.cs
jonasanker: I could do it but this is also just the v1 one method which we have for one month or so :)

TTA777 (approved):

PR #215, opened 2023-10-04
create endpoints to add, update, and deactivate a product

HubertWojcik10:
- issue [here](https://github.com/AnalogIO/shifty-webapp/issues/7)
- I don't know whether I should follow the minimalistic coding approach from AddProduct or the logging approach and many if statements from UpdateProduct
- looking forward to many comments which we can discuss on Thursday


marfavi (dismissed):
Update the branch & format files. The GitHub Actions checks fail on the formatting steps. I believe you can run `dotnet-format /coffeecard` to fix this.

coffeecard/CoffeeCard.WebApi/Controllers/v2/ProductsController.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// Controller for creating, changing, and deactivating a product
+    /// </summary>
+    [ApiController]
+    [Authorize]
+    [ApiVersion("2")]
+    [Route("api/v{version:apiVersion}/products")]
+
+    public class ProductsController : ControllerBase
+    {
+        private readonly ClaimsUtilities _claimsUtilities;
```
TTA777, line 27: This is unused

coffeecard/CoffeeCard.WebApi/Controllers/v2/ProductsController.cs
```diff
@@ -0,0 +1,93 @@
...
+        
+        /// <summary>
+        /// Deactivates a product by setting its visibility to false.
+        /// </summary>
+        /// <param name="productId">The ID of the product to be deactivated.</param>
+        /// <returns>A response indicating the result of the deactivation operation.</returns>
+        /// <response code="200">The request was successful, and the product was deactivated.</response>
+        /// <response code="404">The product with the specified ID was not found.</response>
+        [HttpPost("deactivate/{productId}")]
+        [ProducesResponseType(typeof(void), StatusCodes.Status200OK)]
```
TTA777, line 80: This needs to be limited to the board as well

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/InitiateProductRequest.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+using CoffeeCard.Models.Entities;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product
+{
+    /// <summary>
+    /// Initiate a new product request
+    /// </summary>
+    public class InitiateProductRequest
```
TTA777, line 11: Rename to create Product request. You don't initiate a product

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/InitiateProductResponse.cs
```diff
@@ -0,0 +1,26 @@
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product
+{
+    /// <summary>
+    /// Initiate a new product request
+    /// </summary>
+    public class InitiateProductResponse
```
TTA777, line 10: Initiate -> create

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/InitiateProductRequestWithUserGroups.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using CoffeeCard.Models.Entities;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product;
+
+public class InitiateProductRequestWithUserGroups
```
TTA777, line 6: Initiate -> Create

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,111 @@ public async Task<Product> GetProductAsync(int productId)
...
+            var result = new InitiateProductResponse
+            {
+                Price = product.Price,
+                Description = product.Description,
+                Name = product.Name,
+                NumberOfTickets = product.NumberOfTickets,
+                Visible = product.Visible
+            };
+
+            return Task.FromResult(result);
```
TTA777, line 94: Either remove Task from the return type to make it sequential, or make use of async methods inside it

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,111 @@ public async Task<Product> GetProductAsync(int productId)
...
+                Price = newProduct.Price,
+                Description = newProduct.Description,
+                Name = newProduct.Name,
+                NumberOfTickets = newProduct.NumberOfTickets,
+                Visible = newProduct.Visible
+            };
+            
+            var productUserGroups = allowedUserGroups.Select(userGroup => new ProductUserGroup
+            {
+                ProductId = product.Id,
```
TTA777, line 74: Have you tested product have an Id assigned at this point? It has not been added to the db yet, so likely has no ID

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,111 @@ public async Task<Product> GetProductAsync(int productId)
...
+                Description = product.Description,
+                Name = product.Name,
+                NumberOfTickets = product.NumberOfTickets,
+                Visible = product.Visible
+            };
+
+            return Task.FromResult(result);
+        }
+        
+        private Product GetProduct(int productId)
```
TTA777, line 97: We already have the GetProductASync method, can you use that instead?

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,111 @@ public async Task<Product> GetProductAsync(int productId)
...
+        {
+            return _context.Products.Where(p => p.Id == productId).FirstOrDefault();
+        }
+        
+        public Task<InitiateProductResponse> UpdateProduct(InitiateProductRequest changedProduct)
+        {
+            var product = GetProduct(changedProduct.Id);
+
+            if (changedProduct.Id != default(int))
```
TTA777, line 106: Id will always be set, and never change, so this check doesn't add anything

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,111 @@ public async Task<Product> GetProductAsync(int productId)
...
+            };
+            
+            return Task.FromResult(result);
+        }
+
+        public async Task<bool> DeactivateProduct(int productId)
+        {
+            var product = await _context.Products.FindAsync(productId);
+
+            if (product == null)
```
TTA777, line 154: Typically we throw the EntityNotFoundExceptin when we are unable to find an item, use that instead

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,111 @@ public async Task<Product> GetProductAsync(int productId)
...
+                product.Id = changedProduct.Id;
+            }
+
+            if (changedProduct.NumberOfTickets != default(int))
+            {
+                Log.Information($"Changing NumberOfTickets of product from {product.NumberOfTickets} to {changedProduct.NumberOfTickets}");
+                product.NumberOfTickets = changedProduct.NumberOfTickets;
+            }
+
+            if (!string.IsNullOrEmpty(changedProduct.Name))
```
TTA777, line 118: This would be logged as the name being changed, even if it wasn't. We need to change this condition slightly

TTA777 (changes requested):
Good direction, but changes are needed

coffeecard/CoffeeCard.WebApi/Controllers/v2/ProductsController.cs
TTA777: We might also want to simply use the updateProduct endpoint instead when we change the visibility. Thoughts @jonasanker ?

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
jonasanker: Alternatively, you can just reference the product object and Entity Framework will do the stitching itself

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
```diff
@@ -0,0 +1,26 @@
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product
+{
+    /// <summary>
+    /// Initiate a new product add request
+    /// </summary>
+    public class AddProductRequest
```
jonasanker, line 9: For Swagger/NSwag to provide an example of a request, please also provide code comments
```xml
<example>..</example>
```

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/ProductResponse.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Initiate a new product request
+    /// </summary>
+    public class ProductResponse
+    {
+        [Required]
+        public int Price { get; set; }
+
+        public int NumberOfTickets { get; set; }
```
jonasanker, line 15: Should be required too

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/ProductResponse.cs
```diff
@@ -0,0 +1,26 @@
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product
+{
+    /// <summary>
+    /// Initiate a new product request
+    /// </summary>
+    public class ProductResponse
```
jonasanker, line 10: Add an example code comments for Swagger to provide an response example

coffeecard/CoffeeCard.WebApi/Controllers/v2/ProductsController.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Adds a new product to the database.
+        /// </summary>
+        /// <param name="addProductRequest">The request containing the details of the product to be added and allowed user groups.</param>
+        /// <returns> The newly added product wrapped in a InitiateProductResponse object.</returns>
+        /// <response code="200">The request was successful, and the product was added.</response>
+        [HttpPost("add")]
+        [AuthorizeRoles(UserGroup.Board)]
+        [ProducesResponseType(typeof(ProductResponse), StatusCodes.Status200OK)]
+        public async Task<IActionResult> AddProduct(AddProductRequestWithUserGroups addProductRequest)
```
jonasanker, line 46: Per RESTful design practices, I suggest the API should simply be

```
POST api/v2/products
```

That is remove the add.

Further reference https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#organize-the-api-design-around-resources

coffeecard/CoffeeCard.WebApi/Controllers/v2/ProductsController.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+
+        /// <summary>
+        /// Updates a product with the specified changes.
+        /// </summary>
+        /// <param name="product">The request containing the changes to be applied to the product.</param>
+        /// <returns>A response indicating the result of the update operation.</returns>
+        /// <response code="200">The request was successful, and the product was updated.</response>
+        [HttpPut("update")]
```
jonasanker, line 58: Per restful design practices remove the `update` in the path

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequestWithUserGroups.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using CoffeeCard.Models.Entities;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product;
+
+public class AddProductRequestWithUserGroups
```
jonasanker, line 6: I think this wrapper class is maybe too much, would it not be simple to have a list of EligibleUserGroups in the AddProductRequest since we only use it in this request anyway

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,92 @@ public async Task<Product> GetProductAsync(int productId)
...
 
+        public async Task<ProductResponse> AddProduct(AddProductRequest newProduct, IEnumerable<UserGroup> allowedUserGroups)
+        {
+            var product = new Product()
+            {
+                Price = newProduct.Price,
+                Description = newProduct.Description,
+                Name = newProduct.Name,
+                NumberOfTickets = newProduct.NumberOfTickets,
+                ExperienceWorth = 0,
```
jonasanker, line 69: I'm not keen on setting this to 0 in the repository. Either this should have a default value of 0 on the Controller/API level or it should be provided by the client

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,92 @@ public async Task<Product> GetProductAsync(int productId)
             return product;
         }
 
+        public async Task<ProductResponse> AddProduct(AddProductRequest newProduct, IEnumerable<UserGroup> allowedUserGroups)
+        {
+            var product = new Product()
```
jonasanker, line 63: I think we should check for uniqueness, e.g. check that no such product name already exists to avoid duplicate product names

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        [Required]
+        public int Price { get; set; }
+
+        [Required]
+        public int NumberOfTickets { get; set; }
+
+        [Required]
+        public string Name { get; set; }
```
jonasanker, line 18: Setting Required just ensures that the property is present in the JSON request, not that is has a value. We should add a property to ensure that empty strings are considered invalid. Similarly for Description

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product
+{
+    /// <summary>
+    /// Initiate a new product add request
+    /// </summary>
+    public class AddProductRequest
+    {
+        [Required]
+        public int Price { get; set; }
```
jonasanker, line 12: We should consider a negative integer as an invalid value. Please add a Minumum Data Annotation to check for this. Consider similar checks on NumberofTickets

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/UpdateProductRequest.cs
```diff
@@ -0,0 +1,29 @@
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace CoffeeCard.Models.DataTransferObjects.v2.Product
+{
+    /// <summary>
+    /// Initiate a new product request
+    /// </summary>
+    public class UpdateProductRequest
```
jonasanker, line 9: The DataAnnotations I have suggested for AddProduct should be applied on update as weel

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,92 @@ public async Task<Product> GetProductAsync(int productId)
...
+            };
+
+            return result;
+        }
+
+        public async Task<ProductResponse> UpdateProduct(UpdateProductRequest changedProduct)
+        {
+            var product = await GetProductAsync(changedProduct.Id);
+
+            if (changedProduct.Price != default(int))
```
jonasanker, line 104: What are we checking for here?

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
```diff
@@ -57,6 +58,92 @@ public async Task<Product> GetProductAsync(int productId)
...
+            {
+                Log.Information($"Changing Price of product from {product.NumberOfTickets} to {changedProduct.NumberOfTickets}");
+                product.Price = changedProduct.Price;
+            }
+            if (!string.IsNullOrEmpty(changedProduct.Description))
```
jonasanker, line 109: We could move this check on the Request object using DataAnnotations

jonasanker (dismissed):
First set of comments

coffeecard/CoffeeCard.Library/Services/v2/ProductService.cs
TTA777: @jonasanker Considering, we don't use ExperienceWorth for anything, and will likely remove it entirely in the future, why do we want to expose it in the API, and make the client dependent on it?

jonasanker: Only thought is that by having this on the service/repository layer the behavior of the experiment worth is hidden and has a side effect (that it's always 0). I consider this a technical debt choice.

By giving it a default value on the API layer at least the client has a option to override it

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
HubertWojcik10: I provided code comments but do we really need the examples? It is harder to test it with automatic examples

coffeecard/CoffeeCard.WebApi/Controllers/v2/ProductsController.cs
HubertWojcik10: I removed the update string from the path. However, we don't get the id, we get the whole Product object

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
jonasanker: I'm not sure I understand what you mean with automatic tests. However to give you some more context for my ask, then Swagger is using the examples code documentation block to give a user friendly example to the caller of an API what to expect. See for instance the v2 endpoints here https://core.prd.analogio.dk/swagger/index.html?urls.primaryName=v2#/Vouchers/Vouchers_IssueVouchers

TTA777 (approved):
lgtm

PR #216, opened 2023-10-05
Implement isPerk

marfavi:


coffeecard/CoffeeCard.Library/Utils/ProductUserGroupUtils.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Determines whether a product is a perk or not.
+    /// A product is a perk if it is not available for regular customers.
+    /// </summary>
+    /// <param name="product">The product to check.</param>
+    /// <returns>True if the product is a perk, false otherwise.</returns>
+    public static bool isPerk(Product product)
+    {
+        return product.ProductUserGroup.All(x => x.UserGroup != UserGroup.Customer);
```
jonasanker, line 14: Be aware of Entity Framework defaults to Lazy Loading (See https://learn.microsoft.com/en-us/ef/ef6/querying/related-data#eagerly-loading) that mean that the ProductUserGroup relation is not available but should be loaded

jonasanker (changes requested):

coffeecard/CoffeeCard.Library/Utils/ProductUserGroupUtils.cs
marfavi: Would the latest revision work?

coffeecard/CoffeeCard.Library/Utils/ProductUserGroupUtils.cs
```diff
@@ -0,0 +1,23 @@
...
+public class ProductUserGroupUtils
+{
+    /// <summary>
+    /// Determines whether a product is a perk or not.
+    /// A product is a perk if it is not available for regular customers.
+    /// </summary>
+    /// <param name="context">The DbContext to use for querying the database.</param>
+    /// <param name="product">The product to check.</param>
+    /// <returns>True if the product is a perk, false otherwise.</returns>
+    public static bool isPerk(DbContext context, Product product)
```
jonasanker, line 14: It does not make sense that a utility function should call the database. In the MapperService, the products have already been loaded from the Database, so it would be better to ensure that the User Groups are loaded when fetching the products.

The products are loaded in the `ProductService.GetProducts()` service. I suggest add the include for UserGroups here and then you can filter in your utility method.

jonasanker, line 14: C# has the neat functionality of Extension Methods. I think it makes sense to change your function to become an extension method. See https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods

jonasanker (changes requested):
We are almost there. With the last tweaks, we are good to go :)

PR #224, opened 2023-11-12
Cleanup Products API v2 endpoints

jonasanker:


jonasanker:
@TTA777, @HubertWojcik10 There were some changes in #215 which I think should be cleaned-up thus I have created this PR with these changes. I'm tagging you so you can review them and catch them next time :)
Some of them are a bit oppiniated, others are actually fixes

Highlights
- Example should have no space in them
- Example should have camelCase as this is how our API returns them
- Patch/Put endpoints should have a path parameter /{id} per REST conventions

TTA777:
@jonasanker This has grown quite stale. Is it still something that needs doing, or has it been addressed in separate prs?

PR #233, opened 2023-12-07
217 new admin endpoints: search user

HubertWojcik10:
- create an endpoint to search for user(s) based on their id, name, or email.


A-Guldborg:
I tried to look a bit into the frontend for this :) I found a few things to consider:
- if the "search engine" is good enough, it probably does not need to be able to take multiple / different parameters like we discussed last time. I think the current version is adequate, also not to overload the UI. Might need a discussion once the frontend is added with Omid though.
- If I send an empty string as a parameter, it will cause an error (It changes from empty string to `null`). 
- The API right now both has two contradicting authorization headers (AllowAnonymous and AuthorizeRoles(UserGroup.Board) - probably only should have the latter).
- The ProducesResponseType header for 200 OK is ApiError and not some sort of List/Enumerable/Collection of User.
- The frontend needs to know the total number of items in order for the pagination to work properly, otherwise it will think that there is only `pageNum` elements in total and disables pagination.

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -210,6 +210,38 @@ public async Task UpdateUserGroup(UserGroup userGroup, int userId)
...
+                    users.Add(user);
+                    return users;
+                }
+            }
+
+            users = await _context.Users
+                .Where(u => u.Email.ToLower().StartsWith(search.ToLower()) || u.Name.ToLower().StartsWith(search.ToLower()))
+                .OrderBy(u => u.Id)
+                .Skip(skip).Take(pageLength)
+                .ToListAsync();
```
marfavi, line 233: Greatly reduce complexity of code and eliminate weird edge cases (you may think they're stupid, but they're important) with this change:

```suggestion
            users = await _context.Users
                .Where(u => u.Email.ToLower().StartsWith(search.ToLower()) || u.Name.ToLower().StartsWith(search.ToLower()) || u.Id.ToString().StartsWith(search))
                .OrderBy(u => u.Id)
                .Skip(skip).Take(pageLength)
                .ToListAsync();
```

Alternatively, to enforce an exact match on the ID:
```suggestion
            users = await _context.Users
                .Where(u => u.Email.ToLower().StartsWith(search.ToLower()) || u.Name.ToLower().StartsWith(search.ToLower()) || u.Id.ToString() == search
                .OrderBy(u => u.Id)
                .Skip(skip).Take(pageLength)
                .ToListAsync();
```

marfavi (changes requested):

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
marfavi, line 233: Greatly reduce code complexity and eliminate unintended edge cases (you may think they're stupid, but they're important) with this change:

```suggestion
            var users = await _context.Users
                .Where(u => u.Email.ToLower().StartsWith(search.ToLower()) || u.Name.ToLower().StartsWith(search.ToLower()) || u.Id.ToString().StartsWith(search))
                .OrderBy(u => u.Id)
                .Skip(skip).Take(pageLength)
                .ToListAsync();
```

alternatively, to force exact matches on the id, do this comparison instead:
```
u.Id.ToString() == search
```

marfavi (changes requested):

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
HubertWojcik10: totally agree, I'll try to find some time before Christmas to apply the changes suggested by both of you :)

jonasanker: Also I'm not sure whether we really need pagination. I don't think above suggestions will be that expensive, thus I'm not sure how much we need pagination. :)

jonasanker: Thirdly there is currently no Index on name. We should make sure to add an index to the User column to make the query less expensive

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -210,6 +210,38 @@ public async Task UpdateUserGroup(UserGroup userGroup, int userId)
...
+
+
+            if (users.Count == 0)
```
jonasanker, line 236: Consider using the Any extension method
```dotnet
!users.Count.Any()
```

jonasanker (changes requested):
I have added various comments and added my thoughts to the implementation of the search query. Hope it makes sense else let's start a discussion on Slack :)

Also, we should start implementing tests, also for this API endpoint. I can help to create an example to get you started @HubertWojcik10  :)

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
jonasanker: I was not part of the discussion but would imagine that using Contains over Startswith would give a better experience.

One note on using the Contains method is that it is case sensitive but if you use a SQL `like` method which is case insensitive. To use the like method, you should use the https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbfunctionsextensions.like?view=efcore-8.0

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -201,5 +204,39 @@ private async Task<UserResponse> UserWithRanking(User user)
...
+        /// <response code="401"> Invalid credentials </response>
+        /// <response code="404"> User(s) not found </response>
+        [HttpGet]
+        [AuthorizeRoles(UserGroup.Board)]
+        [AllowAnonymous]
+        [ProducesResponseType(typeof(ApiError), StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(ApiError), StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(ApiError), StatusCodes.Status200OK)]
+        [Route("search")]
+        public async Task<ActionResult<IEnumerable<User>>> SearchUsers([FromQuery] String search, [FromQuery] int pageNum, [FromQuery] int pageLength)
```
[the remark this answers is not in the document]
A-Guldborg: Hmm, I would argue that GET requests should not have bodies if we consider best practice. It definitely is technically possible (I don't know if it is possible with AspNetCore), but considered bad practice. See e.g.:
https://www.baeldung.com/http-get-with-body
https://apidog.com/blog/http-get-request-with-body/

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
A-Guldborg: I don't think we should remove pagination. Pagination is for the frontend - we discussed a solution where the user sees a list of accounts (where they can change user roles) and then a search bar which allows filtering. This means we can avoid having to send a lot of data from the API as we only send `pageLength` accounts.

As for the input length - I don't think we should limit it. We want to show the first `pageLength` accounts on the initial page. I.e., we want to make a "search" for the empty input.

jonasanker: @A-Guldborg Thanks for providing insight into the previous discussion and use case for this new endpoint. I agree that this pagination is relevant

marfavi: @jonasanker An issue with conditionally determining search column is: what if the  user's name is "1" or "1@" (both legal names as of writing)? You wouldn't be able to find user via their name with the queries "1" (would look in the Ids instead) or "1@" (would look in emails instead). Searching in an additional two columns sounds like a good cost of covering these edge cases, considering the expected use frequency of this endpoint

Issue #217, opened 2023-10-19
New Admin Endpoints


jonasanker:
High-level functionality

- [ ] Accounts / Update role https://github.com/AnalogIO/shifty-webapp/issues/6
- [ ] Accounts / Search for user (to be used by shifty)
- [ ] Purchases / Refund https://github.com/AnalogIO/shifty-webapp/issues/8
- [ ] Purchases / Issue Product https://github.com/AnalogIO/shifty-webapp/issues/5
- [x] Products / PUT, POST, DELETE https://github.com/AnalogIO/shifty-webapp/issues/7
- [ ] Statistics / Sum Unused Clips in given time period https://github.com/AnalogIO/shifty-webapp/issues/9

Suggested API design
```
POST api/v2/accounts/search
PATCH api/v2/accounts/{id}/user-group

POST api/v2/products
PUT api/v2/products/{id}
DELETE api/v2/products/{id}

POST api/v2/purchases/issue
PUT api/v2/purchases/{id}/refund

POST api/v2/statistics/unused-clips
```

PR #237, opened 2023-12-07
Adds GET products/all for privileged users

A-Guldborg:
Some weird merge conflicts arised on #232 after #234 was made. Probably due to me branching out from the wrong branch initially. Here is a new pull request instead :)

This adds an endpoint for the product management with frontend in https://github.com/AnalogIO/shifty-webapp/pull/20 and discussed as endpoint in #217 (and slack).

The API will have the endpoint /api/v2/products/all which gives all products, not restricted to a specific usergroup or restricted to only visible products.

This endpoint is only accessible for users with Board privileges and will allow board members to manage the visibilty of products or products that they otherwise don't have access to themselves in the app.

### DIFF
--- a/coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
+++ b/coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
@@ -15,6 +15,7 @@ namespace CoffeeCard.Models.DataTransferObjects.v2.Products
     ///     "name": "Coffee clip card",
     ///     "description": "Coffee clip card of 10 clips",
     ///     "isPerk": true,
+    ///     "visible": true,
     ///     "AllowedUserGroups": ["Manager", "Board"]
     /// }
     /// </example>
--- a/coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
+++ b/coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
@@ -69,6 +69,14 @@ namespace CoffeeCard.Models.DataTransferObjects.v2.Products
         public bool IsPerk { get; set; }
 
         /// <summary>
+        /// Visibility of products for users
+        /// </summary>
+        /// <value>Product visibility</value>
+        /// <example>true</example>
+        [Required]
+        public bool Visible { get; set; }
+
+        /// <summary>
         /// Decides the user groups that can access the product.
         /// </summary>
         /// <value> Product User Groups </value>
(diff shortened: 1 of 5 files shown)


coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
```diff
@@ -68,6 +68,14 @@ public class ProductResponse
         [Required]
         public bool IsPerk { get; set; }
 
+        /// <summary>
+        /// Visibility of products for users
+        /// </summary>
+        /// <value>Product visibility</value>
+        /// <example>true</example>
+        [Required]
+        public bool Visible { get; set; }
```
marfavi, line 77: Add this field to example object at the top of this file.

coffeecard/CoffeeCard.Tests.Unit/Services/v2/ProductServiceTest.cs
```diff
@@ -70,16 +70,14 @@ await productService.UpdateProduct(new UpdateProductRequest()
...
 
-            Assert.Collection<UserGroup>(expected,
-                e => e.Equals(UserGroup.Customer),
-                e => e.Equals(UserGroup.Board));
+            Assert.Collection<ProductUserGroup>(result.ProductUserGroup,
+                e => Assert.Equal(UserGroup.Customer, e.UserGroup),
+                e => Assert.Equal(UserGroup.Board, e.UserGroup));
+
+            Assert.DoesNotContain(UserGroup.Barista, result.ProductUserGroup.Select(e => e.UserGroup));
+            Assert.DoesNotContain(UserGroup.Manager, result.ProductUserGroup.Select(e => e.UserGroup));
```
marfavi, line 80: Doesn't Assert.Collection check that the collection has exactly two elements inside, implying that if it passes, there should be no other elements inside result.ProductUserGroup (making DoesNotContain unnecesary)?

marfavi (changes requested):
Small comments

marfavi (changes requested):
Rename your PR title as well :)

coffeecard/CoffeeCard.Tests.Unit/Services/v2/ProductServiceTest.cs
A-Guldborg: It does, but I have left it there for a reason:

Testing is one of the strongest ways of documenting expected behaviour. Changes in this area in the future should be very aware that it is not just expected that the collection contains those two elements in `Assert.Collection` - but they should have removed the two elements from `Assert.DoesNotContain`. In other words, I believe they convey important information - despite not strictly being necessary in terms of the code test.

marfavi (changes requested):

marfavi (approved):
🚀

PR #240, opened 2023-12-11
Removed public product listing endpoint

### LINKED ISSUE #239: GET Products should always require authorisation
If an expired token is given in the call header for `V2 GET Products`, or if no token is given at all, a default public list of products is returned instead.

This endpoint not requiring authorisation tricks our app into believing that there are no perks available for the user, even though there might be. (Relevant issue https://github.com/AnalogIO/coffeecard_app/issues/564)

Consider the following use case with other endpoints requiring authorisation:
1. App calls endpoint with expired bearer token
2. 401 is returned
3. App intercepts 401 response and ensures a new token is retrieved
4. App calls endpoint with new bearer token
5. Response is succesful

In the case of GET Products, a 401 is not returned when a wrong token is given, so step 2-4 never happens. Instead, a the list of products available to all customers is returned instead (which do not contain any perks at all).

As we do not have a legitimate use case for calling GET Products without an auth token, I propose changing it to always require authorisation instead.

marfavi:
Refactored product retrieval by removing the public listing feature and enforcing authentication. Any unauthenticated access will now result in a 401 Unauthorized response.

Resolves #239, unblocks https://github.com/AnalogIO/coffeecard_app/issues/564

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/IProductService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/IProductService.cs
@@ -8,7 +8,6 @@ namespace CoffeeCard.Library.Services.v2
 {
     public interface IProductService : IDisposable
     {
-        Task<IEnumerable<Product>> GetPublicProductsAsync();
         Task<IEnumerable<Product>> GetProductsForUserAsync(User user);
         Task<Product> GetProductAsync(int productId);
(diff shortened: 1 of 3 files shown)


marfavi:
Can someone generate the new Swagger docs for this patch? (@jonasanker, @TTA777?)

jonasanker (approved):

PR #248, opened 2024-01-18
Adds menu items to products

marfavi:
A clip card can be valid for a range of concrete products (e.g. a cappuccino or a t-shirt). This PR enforces a ticket to specify what concrete product (called a *menu item*) it should be spent on, giving us much more precise statistics on ticket usage.

Endpoint changes:
- **New** `POST /v2/tickets/use/[ticketId]/[menuItemId]` endpoint for using a ticket on a specific menu item
- **Changed all** `/v2/products` endpoints to include a list of eligible menu items in their `ProductResponse` (the field is nullable for backwards compatability reasons).


A-Guldborg:
Will the endpoint for managing these menu items and their relations with products be developed in another PR?

marfavi:
> Will the endpoint for managing these menu items and their relations with products be developed in another PR?

Yes, these are out of scope of this PR

coffeecard/CoffeeCard.Models/Entities/MenuItem.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections.Generic;
+
+namespace CoffeeCard.Models.Entities;
+
+/// <summary>
+/// An item or good that Analog offers, such as a cappuccino or a t-shirt.
+/// </summary>
+public class MenuItem
```
jonasanker, line 8: It should have a uniqueness index to avoid having items with the same name

```suggestion
[Index(nameof(Name), IsUnique = true)]
public class MenuItem
```

coffeecard/CoffeeCard.Models/Entities/MenuItem.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <summary>
+    /// The name of this menu item
+    /// </summary>
+    public string Name { get; set; }
+
+    /// <summary>
+    /// The product(s) that are eligible to redeem this menu item
+    /// </summary>
+    public ICollection<Product> Products { get; set; }
```
jonasanker, line 23: Could the property be named like AssociatedProducts or ProductEligbilities? I think this express the relationship more clearly :)

coffeecard/CoffeeCard.Models/Entities/Product.cs
```diff
@@ -24,6 +24,11 @@ public sealed class Product : IEquatable<Product>
 
         public ICollection<ProductUserGroup> ProductUserGroup { get; set; }
 
+        /// <summary>
+        /// The menu item(s) that this product is eligible to redeem
+        /// </summary>
+        public ICollection<MenuItem> MenuItems { get; set; }
```
jonasanker, line 30: Suggest to call it EligibleMenuItems or similar to clearly express the relationship

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Ticket/TicketResponse.cs
```diff
@@ -57,5 +57,14 @@ public class TicketResponse
...
         [Required]
         public string ProductName { get; set; } = string.Empty;
+
+        /// <summary>
+        /// The name of the menu item that this ticket was used on
+        /// </summary>
+        /// <value>Menu item Id</value>
+        /// <example>Cappuccino</example>
+        /// <remarks>Null if ticket was not used on a menu item</remarks>
+        [JsonProperty(Required = Required.Default)]
```
jonasanker, line 67: I'm not sure we need this annotation but maybe I overlooked something - why did you add it?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
```diff
@@ -16,7 +17,11 @@ namespace CoffeeCard.Models.DataTransferObjects.v2.Products
     ///     "description": "Coffee clip card of 10 clips",
     ///     "isPerk": true,
     ///     "visible": true,
-    ///     "AllowedUserGroups": ["Manager", "Board"]
+    ///     "AllowedUserGroups": ["Manager", "Board"],
```
jonasanker, line 20: Should be camelCase :)

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
```diff
@@ -84,5 +89,13 @@ public class ProductResponse
         /// <example> Manager, Board </example>
         [Required]
         public IEnumerable<UserGroup> AllowedUserGroups { get; set; } = new List<UserGroup>();
```
jonasanker, line 91: I suggest not to add a default value despite the compiler warning. When we move to .NET 8, we will introduce the required property which I believe is a better solution when we do want to it to have a default value. This also applies for MenuItems

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
```diff
@@ -84,5 +89,13 @@ public class ProductResponse
...
+
+        /// <summary>
+        /// The menu items that this product can be used on.
+        /// </summary>
+        /// <value>Menu items</value>
+        /// <example>Cappuccino, Caffe Latte</example>
+        /// <remarks>Optional for backwards compatibility</remarks>
+        public IEnumerable<MenuItemResponse> MenuItems { get; set; } = new List<MenuItemResponse>();
```
jonasanker, line 99: Suggest to rename to EligibleMenuItems to clearly express the intent

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/MenuItemResponse.cs
```diff
@@ -0,0 +1,32 @@
...
+        [Required]
+        public int Id { get; set; }
+
+        /// <summary>
+        /// Name of menu item
+        /// </summary>
+        /// <value>Menu item name</value>
+        /// <example>Cappuccino</example>
+        [Required]
+        public string Name { get; set; } = string.Empty;
```
jonasanker, line 30: We should not have a default value despite the compiler warning

coffeecard/CoffeeCard.WebApi/Controllers/v2/TicketsController.cs
```diff
@@ -50,5 +47,27 @@ public async Task<ActionResult<List<TicketResponse>>> Get([FromQuery] bool inclu
...
+        /// <response code="200">Successful request</response>
+        /// <response code="401">Invalid credentials</response>
+        /// <response code="403">Menu item cannot be redeemed by this ticket</response>
+        /// <response code="404">No ticket on account or menu item not found</response>
+        [ProducesResponseType(typeof(UsedTicketResponse), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(typeof(MessageResponseDto), StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(MessageResponseDto), StatusCodes.Status404NotFound)]
+        [HttpPost("use/{ticketId:int}/{menuItemId:int}")]
+        public async Task<ActionResult<UsedTicketResponse>> UseTicket(int ticketId, int menuItemId)
```
jonasanker, line 66: I like your design idea to utilize route/path parameters but I believe this does not read as one would normally expect for REST APIs. You use the ticketId but in fact the TicketService takes a ProductId and then the MenuItem Id. Due to this I think this becomes a bit confusing when have the endpoint like

```
POST tickets/use/{productId}/{menuItemId} (Pull Request)
```

Usually you would expect that we would point to a specific ticketId like
```
POST tickets/{id}/use ..
```
but since we use the productId for user friendlyness, I believe this does not read well.

I'm trying to find a good suggestion how you could make this more inline with REST practices (there's no standard though) which would read like the sentence _Use one ticket for a given product for a given menu item_. I can't think of a good design with paths, so my best suggestion is as we did before

```
POST tickets/use
{
   productId: 1,
   useForMenuItem: 1
}
```

jonasanker: Also, if using Route Parameters, then lets use the `FromRoute` annotation for clarity like

```suggetion
public async Task<ActionResult<UsedTicketResponse>> UseTicket([FromRoute] int ticketId, [FromRoute] int menuItemId)
```

coffeecard/CoffeeCard.WebApi/Controllers/v2/TicketsController.cs
```diff
@@ -50,5 +47,27 @@ public async Task<ActionResult<List<TicketResponse>>> Get([FromQuery] bool inclu
...
+        {
+            var user = await _claimsUtilities.ValidateAndReturnUserFromClaimAsync(User.Claims);
+
+            return Ok(await _ticketService.UseTicketAsync(user, ticketId, menuItemId));
```
jonasanker, line 70: UseTicketAsync does not take a TicketId but a ProductId

coffeecard/CoffeeCard.Library/Services/v2/TicketService.cs
```diff
@@ -46,14 +45,18 @@ public async Task IssueTickets(Purchase purchase)
...
+                .Include(t => t.UsedOnMenuItem)
                 .Select(t => new TicketResponse
                 {
                     Id = t.Id,
                     DateCreated = t.DateCreated,
                     DateUsed = t.DateUsed,
                     ProductId = t.ProductId,
-                    ProductName = t.Purchase.ProductName
+                    ProductName = t.Purchase.ProductName,
+                    UsedOnMenuItemName = t.UsedOnMenuItem != null ? t.UsedOnMenuItem.Name : null
```
jonasanker, line 59: Suggest to use the [Null Conditional Operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)

```suggestion
UsedOnMenuItemName = t.UsedOnMenuItem?.Name
```

coffeecard/CoffeeCard.Library/Services/v2/TicketService.cs
```diff
@@ -83,6 +86,35 @@ public async Task<UsedTicketResponse> UseTicketAsync(User user, int productId)
             };
         }
 
+        public async Task<UsedTicketResponse> UseTicketAsync(User user, int productId, int menuItemId)
```
jonasanker, line 89: I believe it would be good to check to check if the ProductId is valid (that it exists) at first in the method before querying for tickets or menu items

jonasanker (changes requested):
Overall looks good, see my suggestions to provide more clear properties and rework on the REST design :)

coffeecard/CoffeeCard.Library/Services/v2/TicketService.cs
marfavi: I tried that, but it doesn't compile:

> An expression tree lambda may not contain a null propagating operator. [(CS8072)](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/expression-tree-restrictions?f1url=?appId=roslyn&k=k(CS8072))

Oddly enough dropping the question mark _does_ compile, however I would not do this.

jonasanker: Alright, let's close this point then :)

coffeecard/CoffeeCard.Library/Services/v2/TicketService.cs
marfavi: Addressed in 4011699489e03cefe6b6666037d7c3262bc97abd

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Ticket/TicketResponse.cs
marfavi: Oversight - the Default value is, well, the default anyway. At some point it was set to DisallowNull instead ("The property is not required but it cannot be a null value.") I'm still not sure if that is needed though, what do you think?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
marfavi: Addressed in 3dee1ab727bad46b91e42622e3b19b97b7b92bf8

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/MenuItemResponse.cs
marfavi: Addressed in 3dee1ab727bad46b91e42622e3b19b97b7b92bf8

coffeecard/CoffeeCard.Models/Entities/MenuItem.cs
marfavi: Addressed in fe39ffed2d7d416aee95f353586cf80e70afeb6f

coffeecard/CoffeeCard.Models/Entities/Product.cs
marfavi: Addressed in fe39ffed2d7d416aee95f353586cf80e70afeb6f

coffeecard/CoffeeCard.Models/Entities/MenuItem.cs
marfavi: Addressed in fe39ffed2d7d416aee95f353586cf80e70afeb6f

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
marfavi: Addressed in fe39ffed2d7d416aee95f353586cf80e70afeb6f

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Products/ProductResponse.cs
marfavi: Addressed in fe39ffed2d7d416aee95f353586cf80e70afeb6f

coffeecard/CoffeeCard.WebApi/Controllers/v2/TicketsController.cs
marfavi: Addressed in e93b3e189f2a16dde9c1b6eb34a2b6ab52fa58f2

coffeecard/CoffeeCard.WebApi/Controllers/v2/TicketsController.cs
marfavi: Addressed in 15a7eec0f68f80a50105d0ffa8e04022c4589783

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Ticket/TicketResponse.cs
marfavi: Removed in 7dd4783a594872f97a7592461f2857f59403ead5

jonasanker (dismissed):
🌮 🎉

jonasanker (approved):

PR #249, opened 2024-01-18
Search returns single element that contains list

A-Guldborg:
See below screenshot showing that the service returns a `UserSearchResponse` and not an enumerable.
image

This affects Shifty when we use the GenerateApi tool and the generated code expects to return a list.

A `UserSearchResponse` contains an enumerable of SimpleUserResponse and a TotalUsers (count) property.


jonasanker:
Good catch. Sorry for being too quick

jonasanker (approved):

PR #251, opened 2024-01-20
Admin endpoints for menu items

A-Guldborg:
Adds admin endpoints to #248 

It is possible to add, update and read menu items as a board member.

Board members can also specify menu items on products when adding and updating products.

@marfavi consider if you want this in your feature-branch or if this should be merged to main *after* feature-branch is merged 😄


coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/UpdateProductRequest.cs
```diff
@@ -80,5 +81,13 @@ public class UpdateProductRequest
...
         [Required]
         public IEnumerable<UserGroup> AllowedUserGroups { get; set; } = new List<UserGroup>();
+
+        /// <summary>
+        /// Gets or sets the eligible menu items for the product.
+        /// </summary>
+        /// <value> Product Menu Item Ids </value>
+        /// <example> 1, 2 </example>
+        [Required]
+        public IEnumerable<int> MenuItemIds { get; set; } = new List<int>();
```
jonasanker, line 91: Do we need the default value?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/UpdateMenuItemRequest.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </example>
+    public class UpdateMenuItemRequest
+    {
+        /// <summary>
+        /// Gets or sets the ID of the product to update.
+        /// </summary>
+        /// <value> Product Id </value>
+        /// <example> 1 </example>
+        [Required]
+        public int Id { get; set; }
```
jonasanker, line 25: If we change to an API design like `PUT menuitems/[id]`, this property is not required

coffeecard/CoffeeCard.Library/Services/v2/IMenuItemService.cs
```diff
@@ -0,0 +1,15 @@
...
+using CoffeeCard.Models.DataTransferObjects.v2.Product;
+using CoffeeCard.Models.DataTransferObjects.v2.Products;
+
+namespace CoffeeCard.Library.Services.v2
+{
+    public interface IMenuItemService : IDisposable
+    {
+        Task<IEnumerable<MenuItemResponse>> GetAllMenuItemsAsync();
+        Task<MenuItemResponse> AddMenuItem(AddMenuItemRequest newMenuItem);
+        Task<ChangedMenuItemResponse> UpdateMenuItem(UpdateMenuItemRequest changedMenuItem);
```
jonasanker, line 13: Wonder if we can just reuse the `MenuItemResponse`?

coffeecard/CoffeeCard.Library/Services/v2/MenuItemService.cs
```diff
@@ -0,0 +1,103 @@
...
+                .Select(p => new MenuItemResponse
+                {
+                    Id = p.Id,
+                    Name = p.Name
+                })
+                .ToListAsync();
+        }
+
+
+        private async Task<bool> CheckMenuItemUniquenessAsync(string name, int id = 0)
```
jonasanker, line 37: I'm not sure I can see in what scenario the `id` would be `0`. Suggest to check by the name of a menu item

coffeecard/CoffeeCard.Library/Services/v2/MenuItemService.cs
```diff
@@ -0,0 +1,103 @@
...
+            var result = new MenuItemResponse
+            {
+                Id = menuItem.Id,
+                Name = menuItem.Name
+            };
+
+            return result;
+        }
+
+        public async Task<ChangedMenuItemResponse> UpdateMenuItem(UpdateMenuItemRequest changedMenuItem)
```
jonasanker, line 70: I would suggest to change the order of operations in this method
1. Check that Id exists
2. Check for uniqueness of new name
3. Perform changes

jonasanker (changes requested):
A few changes requested

coffeecard/CoffeeCard.Library/Services/v2/MenuItemService.cs
A-Guldborg: Hmm, I think I somewhat over-engineered this part. My thought was along the lines: If I want to update a menu item, I want to check that I am not saving it as an existing name for a different menu item - but it should not throw an exception if I try to "save" it with the same name. This would make sense if a menu item has more properties - but since we only have the name property, we do not need it. We can just handle it in the frontend (don't save if we do not make changes), and then change this method in the future when (and if) we need it.

If you agree I'll remove the id part of this method, otherwise I suggest making an extra method such that we have a method checking for uniqueness for new menu item (does not take id into account) and for updating menu items (taking id into account).

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
```diff
@@ -15,7 +15,8 @@ namespace CoffeeCard.Models.DataTransferObjects.v2.Product
     ///     "NumberOfTickets": 10,
     ///     "Description": "xxx",
     ///     "Visible": true,
-    ///     "AllowedUserGroups": ["Manager", "Board"]
+    ///     "AllowedUserGroups": ["Manager", "Board"],
+    ///     "MenuItemIds": [1, 2]
```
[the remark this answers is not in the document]
A-Guldborg: I have fixed this, but all other properties on our products are PascalCase. Let me know if you want me to make the rest of our products camelcase as well as part of this PR?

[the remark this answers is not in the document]
jonasanker: The properties are converted to camelCase by nswag thus while the properties in code are PascalCase, the properties in json are camelCase 😉

Therefore our examples should be in camelCase 🙌🏻

coffeecard/CoffeeCard.Library/Services/v2/MenuItemService.cs
jonasanker: I don't see us hit that scenario at this time so suggest to remove the id

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Product/AddProductRequest.cs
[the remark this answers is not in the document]
A-Guldborg: I have fixed this for the `AddMenuItemRequest`, `UpdateMenuItemRequest` and `UpdateProductRequest` in this commit: https://github.com/AnalogIO/analog-core/pull/251/commits/49e892f8ceb83ecf1005a53de52cd71338037d1d 
The `AddProductRequest` (where your comment was added) is addressed in https://github.com/AnalogIO/analog-core/pull/251/commits/b3ba9a163eb1aabbbb75149c91057bcb79d32522 

Can you please verify it has been fixed as expected?

jonasanker (approved):

PR #252, opened 2024-01-20
Adds V2 GET products/[id] endpoint

marfavi:
Required for the app to know what menu items can be claimed with a ticket (if the ticket is for a product that is no longer for sale)


jonasanker (approved):
Lgtm. Cheers 🍻

PR #264, opened 2024-02-20
Webhook for bulk updating accounts usergroup

A-Guldborg:
Implements the endpoint `PUT /api/v2/webhooks/accounts/user-group`,
allowing an external service to update user groups in bulk.

Example request body:
```
{
   "privilegedUsers": [
       { "accountId": 123, "userGroup": "Manager" },
       { "accountId": 124, "userGroup": "Barista" }
    ]
}
```

The X-Api-Key header should be present on the request with a valid API
key.

The userGroup should be one of:
- Barista
- Manager
- Board
- (Customer will also work, but will have the same effect as not
specifying the user in the request)

### Notes on testing

I have been unable to implement an integration test, since
EFCore.InMemory does not seem to support the performance-optimized
ExecuteUpdate that is implemented in EF Core 7 and used in this service.
Reference:
https://stackoverflow.com/questions/74907256/ef-7-new-executedelete-and-executeupdate-methods-not-working-on-an-in-memory-d

### Business logic

Initially, business logic was implemented using client-side hashing to verify duplicate requests did not use many resources. Later, it was changed to server-side hashing however hashing was completely removed again, as it would cause weird behaviour in combination with the below mentioned existing endpoint. Now it will be clearer that the state in the external system will correct itself if analog-core and the external system are out of sync.

### Notes on affected existing endpoints

This change will also mark our `PATCH /api/v2/account/<id>/user-group`
endpoint, as any calls to the new endpoint will override any user-groups
set by the old patch endpoint.
The endpoint is used by the Shifty User Manager from https://github.com/AnalogIO/shifty-webapp/pull/21

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
@@ -323,7 +323,7 @@ namespace CoffeeCard.Library.Services.v2
 
                 if (hash.Equals(cachedHash))
                 {
-                    Log.Information($"Duplicate update usergroup request detected: {hash}");
+                    Log.Information("Duplicate update user group request detected: {hash}", hash);
                     return;
                 }
             }
--- a/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
@@ -9,23 +9,28 @@ using CoffeeCard.Models.DataTransferObjects.v2.User;
 using CoffeeCard.Models.Entities;
 using Microsoft.AspNetCore.Http;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Caching.Memory;
 using Serilog;
 
 namespace CoffeeCard.Library.Services.v2
 {
     public class AccountService : IAccountService
     {
+        private const string AccountsWebhooksHashCacheKey = "AccountsWebhooksHash";
+
         private readonly CoffeeCardContext _context;
         private readonly IEmailService _emailService;
         private readonly IHashService _hashService;
         private readonly ITokenService _tokenService;
+        private readonly IMemoryCache _memoryCache;
 
-        public AccountService(CoffeeCardContext context, ITokenService tokenService, IEmailService emailService, IHashService hashService)
+        public AccountService(CoffeeCardContext context, ITokenService tokenService, IEmailService emailService, IHashService hashService, IMemoryCache memoryCache)
         {
             _context = context;
             _tokenService = tokenService;
             _emailService = emailService;
             _hashService = hashService;
+            _memoryCache = memoryCache;
         }
 
         public async Task<User> RegisterAccountAsync(string name, string email, string password, int programme)
@@ -274,6 +279,39 @@ namespace CoffeeCard.Library.Services.v2
         private static string EscapeName(string name)
         {
             return name.Trim('<', '>', '{', '}');
+        }
+
+        public async Task UpdatePriviligedUserGroups(WebhookUpdateUserGroupRequest request)
+        {
+            if (!_memoryCache.TryGetValue(AccountsWebhooksHashCacheKey, out string cachedHash))
+            {
+                cachedHash = request.Hash;
+
+                var cacheExpiryOptions = new MemoryCacheEntryOptions
+                {
+                    AbsoluteExpiration = DateTime.UtcNow.AddHours(48),
+                    SlidingExpiration = TimeSpan.FromHours(24)
+                };
+
+                Log.Information("Set {AccountsWebhooksHashCacheKey} in Cache", AccountsWebhooksHashCacheKey);
+                _memoryCache.Set(AccountsWebhooksHashCacheKey, cachedHash, cacheExpiryOptions);
+            }
+
+            if (cachedHash.Equals(request.Hash)) {
+                return;
+            }
+
+            // TODO: Fix Bulk Update in EFCore 6
+            await _context.Users
+                .Where(u => u.UserGroup != UserGroup.Customer)
+                .ExecuteUpdateAsync(u => u.SetProperty(u => u.UserGroup, UserGroup.Customer));
+
+            foreach (var item in request.PrivilegedUsers)
+            {
+                await _context.Users
+                    .Where(u => u.Id == item.AccountId)
+                    .ExecuteUpdateAsync(u => u.SetProperty(u => u.UserGroup, item.UserGroup));
+            }
         }
     }
 }
--- a/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
@@ -297,7 +297,7 @@ namespace CoffeeCard.Library.Services.v2
 
             if (request.Hash.Equals(cachedHash))
             {
-                Log.Information($"Hash seen before: {request.Hash}");
+                Log.Information($"Duplicate update usergroup request detected: {request.Hash}");
                 return;
             }
 
--- a/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
@@ -285,8 +285,6 @@ namespace CoffeeCard.Library.Services.v2
         {
             if (!_memoryCache.TryGetValue(AccountsWebhooksHashCacheKey, out string cachedHash))
             {
-                cachedHash = request.Hash;
-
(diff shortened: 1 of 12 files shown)


coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -275,5 +280,37 @@ private static string EscapeName(string name)
...
+                    SlidingExpiration = TimeSpan.FromHours(24)
+                };
+
+                Log.Information("Set {AccountsWebhooksHashCacheKey} in Cache", AccountsWebhooksHashCacheKey);
+                _memoryCache.Set(AccountsWebhooksHashCacheKey, request.Hash, cacheExpiryOptions);
+            }
+
+            if (request.Hash.Equals(cachedHash))
+            {
+                Log.Information($"Hash seen before: {request.Hash}");
```
jonasanker, line 300: I don't think we need this log statement. If you think we do, then maybe find a more descriptive text

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/WebhookUpdateUserGroupRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Represents a request to update user groups in bulk
+    /// </summary>
+    public class WebhookUpdateUserGroupRequest
+    {
+        /// <summary>
+        /// The hash of the request
+        /// </summary>
+        public string Hash { get; set; }
```
jonasanker, line 14: After a second thought, do we need to take the hash as input? Maybe an alternative solution would be that the backend calculates the hash it self based on the list received? It would give the same effect for the backend but we relieve the client for this task.

jonasanker (changes requested):
A few comments. 

I think it still makes sense to keep the existing endpoint in the case you need to do an temporary override, then the user manager could still be used.

Such a shame with testing. I think we should create an issue on this so we can solve this in the future. Possibly replacing the in memory database?

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
A-Guldborg: I have updated this. In fact, I had it locally but forgot to commit it at midnight 😢 

I want to keep some logging for duplicate requests as I think it can be useful in cases where the board approaches us with the question "why is X happening?" then we have at least some debugging information without having access to their implementation 😅

A-Guldborg:
Hi @jonasanker , it is now ready for re-review based on my above comments/replies to your change-requests.

Please note that I have not added the index that you mentioned in-person. Do you still reckon we should have an index on e.g. UserGroup or something else? 😊

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -275,5 +282,62 @@ private static string EscapeName(string name)
...
+                // and format each one as a hexadecimal string.
+                for (int i = 0; i < data.Length; i++)
+                {
+                    sBuilder.Append(data[i].ToString("x2"));
+                }
+            }
+
+            // Return the hexadecimal string.
+            return sBuilder.ToString();
+        }
```
jonasanker, line 304: What are your thoughts on this implementation vs overriding e.g. hashcode or equals function to calculate a deterministic hash?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/WebhookUpdateUserGroupRequest.cs
jonasanker: I agree the use case for the hash is very slim, to an extend where I'm considering to scratch it entirely? What we are really implementing with the hash is just a prevention for multiple concurrent calls spamming the database. This is probably a rare scenario regardless.

Not Modified would be the most appropriate in the case where we do not modify the database.

I believe it is better to maintain the endpoint than removing. E.g. lets say the client implementation calling the webhook is down, then it is more desirable to have an endpoint than the board editing in the database directly, thus I suggest to keep it.

You have though a good point on how the cache can limit this. I therefore either suggest to 1) remove the cache, 2) lower the cache expiration or 3) maybe pull a member list from the database and compare with the requested object. We only then update the database when the request is not equals to what is stored in the DB.
With an index on UserGroup, this would not be a heavy database operation.

jonasanker (changes requested):

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/WebhookUpdateUserGroupRequest.cs
A-Guldborg: I think removing the cache completely is the best solution. If the server sends a new request there is likely a reason for it. If it is just a daily CRON-job or similar trigger action, then we should probably still make sure that the database is still "aligned". We could discuss with the board how often they want this solution to sync. If they want it based on *any* member info is updated, then we need some sort of hash I would say (consider the case where they are updating profiles just after accepting new baristas, then they will checking the list a lot). If they want it to be based on a CRON job, e.g. daily, we could either disregard the cache or instead just check "time since last request" and verify it is at least X hours old?

I quickly discussed this with Omid today, and I think we kinda thought a daily request before the café opens but after the server capacity increases could work? (probably doesn't require the server capacity to be scaled up so can also run in the middle of the night I guess).

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
A-Guldborg: I am uncertain in which cases this would not be deterministic? 😅
Are you thinking of the scenario where the list is sent in a new order? In that case I would argue it is a new request - for one reason or another, the requester/system has actively sent a new request. We could also fix this by simply sorting the list (e.g. by id) before hashing it.

If you mean equals function such that two request objects are equal if requestA.length == requestB.length && requestA.union(requestB) == requestA.length (or a similar solution comparing the two objects contents with each other), then we would need to cache the whole request object and not just a hash value.

However please read my above comment on my take on the overall value of hashing.

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -275,5 +282,62 @@ private static string EscapeName(string name)
...
+                        SlidingExpiration = TimeSpan.FromHours(24)
+                    };
+
+                    Log.Information("Set {AccountsWebhooksHashCacheKey} in Cache", AccountsWebhooksHashCacheKey);
+                    _memoryCache.Set(AccountsWebhooksHashCacheKey, hash, cacheExpiryOptions);
+                }
+
+                if (hash.Equals(cachedHash))
+                {
+                    Log.Information($"Duplicate update usergroup request detected: {hash}");
```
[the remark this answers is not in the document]
A-Guldborg: For reference, I edited your suggestion before committing it in order to keep the indent.

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
jonasanker: Closing this discussion

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/WebhookUpdateUserGroupRequest.cs
jonasanker: Lets remove the hash entirely, add the index and ship it :)

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/WebhookUpdateUserGroupRequest.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Represents a request to update user groups in bulk
+    /// </summary>
+    public class WebhookUpdateUserGroupRequest
+    {
+        /// <summary>
+        /// List of accounts and their new user groups
+        /// </summary>
+        public IEnumerable<AccountUserGroup> PrivilegedUsers { get; set; }
```
jonasanker, line 17: Required

coffeecard/CoffeeCard.Models/Entities/AccountUserGroup.cs
```diff
@@ -0,0 +1,21 @@
+namespace CoffeeCard.Models.Entities;
+
+/// <summary>
+/// Represents an account user group update
+/// </summary>
+public class AccountUserGroup
```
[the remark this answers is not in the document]
jonasanker: Properties should be required

jonasanker (changes requested):

jonasanker (approved):

PR #266, opened 2024-02-25
Fix/open api spec

TTA777:
- Changed json serilizer from newtonsoft 
- Removed state from MobilePayPaymentDetails

PR #267, opened 2024-02-28
Changes Users to property

A-Guldborg:
With the new Json serializer from #266, fields are no longer serialized.

Preferably, we would have added the Json options `options.JsonSerializerOptions.IncludeFields = true;` to ensure that no other functionality broke. However, due to an issue raised [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2423), the schema in Swagger will not show fields despite setting this option. Thus, our API-generation would not work for e.g. Shifty that uses this.

Instead, I have looked through `CoffeeCard.Models` and not found other examples where we are using fields and not properties.


jonasanker (approved):
Lgtm

PR #268, opened 2024-03-11
Introduce Feature Flag to control MobilePay Webhooks registration

jonasanker:
When migrating to the new MobilePay Vipps API, the existing MobilePay Webhooks API will be disabled, thus we need a way to disable the feature until the new integration has been developed. We use Feature Flags to disable registration to the Webhooks API.

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/WebhookService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/WebhookService.cs
@@ -26,7 +26,6 @@ namespace CoffeeCard.Library.Services.v2
         private readonly IMobilePayWebhooksService _mobilePayWebhooksService;
         private readonly MobilePaySettingsV2 _mobilePaySettings;
         private readonly IMemoryCache _memoryCache;
-        private readonly IFeatureManager _featureManager;
 
         public WebhookService(CoffeeCardContext context, IMobilePayWebhooksService mobilePayWebhooksService,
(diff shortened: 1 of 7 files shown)


coffeecard/CoffeeCard.Library/Services/v2/WebhookService.cs
```diff
@@ -25,6 +26,7 @@ public class WebhookService : IWebhookService
         private readonly IMobilePayWebhooksService _mobilePayWebhooksService;
         private readonly MobilePaySettingsV2 _mobilePaySettings;
         private readonly IMemoryCache _memoryCache;
+        private readonly IFeatureManager _featureManager;
```
TTA777, line 29: This is unused

coffeecard/CoffeeCard.Library/Services/v2/WebhookService.cs
```diff
@@ -86,16 +88,6 @@ private async Task SyncWebhook(WebhookConfiguration webhook)
                     await DisableAndRegisterNewWebhook(webhook);
                     return;
                 }
-
```
TTA777, line 89: Maybe I'm missing something, but isn't this the bit that's supposed be feature toggled?

TTA777 (commented):
Two things I'd like cleared up, see comments

TTA777 (approved):

PR #278, opened 2024-04-15
Changes unused clips response type to enumerable

A-Guldborg:
Currently, auto-generated swagger documentation shows only a single response item in the example response value, when the actual result is a JSON array (one element per product type for the given period).

This fix will prevent issues for developers in shifty when generating the API interface.


jonasanker (approved):
Lgtm

PR #284, opened 2024-09-24
RefundPayment on PurchasesController

fredpetersen:
This hasn't been tested, as it is currently impossible
See https://developer.mobilepay.dk/docs/app-payments/transition-to-one-platform#test


marfavi (approved):
Pengvinos, take the wheel 🚀

marfavi (approved):

PR #287, opened 2024-10-01
feat: adds ticket status (refunded) & adds refund logic

marfavi:
- Tickets are now marked as refunded when calling `PurchaseService.RefundPurchase`.

- Enhanced refund logic to ensure only fully unused purchases are refundable.

- Replaced the boolean `IsUsed` property with a `Status` property on tickets to better represent 3 states (unused, used, refunded).

- Added `IsConsumed` and `IsConsumable` read-only convenience properties to Ticket.

- Improved method clarity by renaming operations related to unused tickets.


coffeecard/CoffeeCard.Models/Entities/Ticket.cs
```diff
@@ -13,7 +13,10 @@ public class Ticket
 
         public int ProductId { get; set; }
 
-        public bool IsUsed { get; set; }
```
TTA777, line 16: On second thought, I think we should play this safe and not delete anything until. Mark it as obsolete (true) for now. That way, nothing can reference it without breaking the build, but we preserve the data

TTA777 (changes requested):

marfavi:
I have no idea why build fails. Seems like some generated files still attempt to access the obsolete `IsUsed` field @TTA777

TTA777:
@marfavi Yeah, that would be my fancy source generator for builders we use for writing the integration tests. They don't account for properties being obsolete. I can push a fix for it later

TTA777 (dismissed):
Ship it

TTA777 (approved):
Ship it now

PR #289, opened 2024-10-22
Login with magic link

A-Guldborg:
This feature aims at enabling magic link authentication through mails. The mail will then provide a deeplink for the correct application (app/shifty) on valid login and use refresh tokens to stay logged in.

This is only for Shifty at this point, since the implementation is missing for the app frontend. Once implemented in both places, we should do a soft roll-over of requiring users to update the app, possibly with a hard requirement at the beginning of the new semester.


A-Guldborg:
Email service is going to be hard to test compared to the testing probably required, considering it is mostly building templates by replacing structured values. This is due to it opening files which is different between running tests locally and in our workflow.

This leaves us slightly below required testing coverage requirements but imo it should be okay in this very case.

Other than that, ready for review :)

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/UserLoginResponse.cs
```diff
@@ -0,0 +1,24 @@
+namespace CoffeeCard.Models.DataTransferObjects.v2.User
+{
+    /// <summary>
+    /// User login response object
+    /// </summary>
+    /// <example>
+    /// {
+    ///     "jwt": "hidden",
+    ///     "refreshToken": "hidden"
```
fredpetersen, line 9: I think we usually use `"[no example given]"` instead of "hidden"

fredpetersen (dismissed):

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -289,5 +300,44 @@ await _context.Users
...
+            if (user is null)
+            {
+                // Should not throw error to prevent showing a malicious user if an email is already registered
+                return;
+            }
+            var magicLinkTokenHash = _tokenServiceV2.GenerateMagicLink(user);
+            await _emailServiceV2.SendMagicLink(user, magicLinkTokenHash, loginType);
+        }
+
+        public async Task<UserLoginResponse> GenerateTokenPair(string token)
```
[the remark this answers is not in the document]
A-Guldborg: Hmm, in my opinion `UserLoginResponse` is a data object representing/carrying a token pair. But on the other hand, token might be a bit overloaded / generic for this use-case so I have renamed it. Hope you like the new name, otherwise please suggest alternatives :)

coffeecard/CoffeeCard.Models/Entities/Token.cs
```diff
@@ -3,31 +3,76 @@
 
 namespace CoffeeCard.Models.Entities
 {
-    public class Token
```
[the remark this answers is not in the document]
A-Guldborg: Done in https://github.com/AnalogIO/analog-core/pull/289/commits/3cac5a86d2c2b70db2bae4fb0c9a223d97157e19

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/UserLoginResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// }
+    /// </example>
+    public class UserLoginResponse
```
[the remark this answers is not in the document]
A-Guldborg: Done in https://github.com/AnalogIO/analog-core/pull/289/commits/c175b82c8e42d5a0276211d1b8b585fc0d338862

coffeecard/CoffeeCard.Common/Configuration/EnvironmentSettings.cs
```diff
@@ -11,6 +11,8 @@ public class EnvironmentSettings : IValidatable
 
         [Required] public string DeploymentUrl { get; set; }
 
+        [Required] public string ShiftyUrl { get; set; }
```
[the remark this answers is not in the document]
A-Guldborg: @duckth can you help with this part? I am not quite sure how the deployment works :)

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/User/UserLoginResponse.cs
A-Guldborg: Done in https://github.com/AnalogIO/analog-core/pull/289/commits/c175b82c8e42d5a0276211d1b8b585fc0d338862

A-Guldborg:
@TTA777 while attempting to use the different test data builders; I do see the benefits of using a builder, but it does not seem to allow us to use parameterized constructors which we define for the `Token` type in this PR to ensure a TokenHash and Type is set. We are for instance setting this in the constructor to ensure we can call `type.getExpiresAt()` to set the default expiration value for the Token, and still have it overridable in case this is needed somewhere. For example, we use this in the tests for getting a refresh token using an expired token.

Currently getting the following error message on tests using the TokenBuilder:
>    System.MissingMethodException : Cannot dynamically create an instance of type 'CoffeeCard.Models.Entities.Token'. Reason: No parameterless constructor defined.

How do you reckon we go about this? I can't see any "pretty" solution for it. Ideas:
- We make it parameterless but explicitly set the expires at after every time we create the object
- We (i guess you) investigate if it is possible to make the builder use the defined constructor (I have no clue how the builders generally work... hehe)
- We make the expires property into a method calling the tokentype's static method for default expiration period but also being aware of its token's created timedate to calculate the expires / expired value. This however means we cannot set the value explicitly in the tests and will have to find a way to instead reset / change the "clock" in the tests to simulate being past the token's expiration.

These are my initial three ideas, perhaps you have another idea that is cleaner but I don't think any of the above are very clean..

I will look into integration test at some other time :)

coffeecard/CoffeeCard.Common/Configuration/EnvironmentSettings.cs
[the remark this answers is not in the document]
duckth: Yes

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
```diff
@@ -0,0 +1,66 @@
...
+    private readonly CoffeeCardContext _context;
+    private readonly IHashService _hashService;
+
+    public TokenService(CoffeeCardContext context, IHashService hashService)
+    {
+        _context = context;
+        _hashService = hashService;
+    }
+
+    public async Task<string> GenerateMagicLink(User user)
```
TTA777, line 22: Naming issue, this does not generate a magic link, but a token

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
```diff
@@ -0,0 +1,66 @@
...
+        var hashedToken = _hashService.Hash(refreshToken);
+        _context.Tokens.Add(new Token(hashedToken, TokenType.Refresh) { UserId = user.Id });
+        await _context.SaveChangesAsync();
+        return refreshToken;
+    }
+
+    public async Task<Token> GetValidTokenByHashAsync(string tokenString)
+    {
+        var tokenHash = _hashService.Hash(tokenString);
+        var foundToken = await _context.Tokens.Include(t => t.User).FirstOrDefaultAsync(t => t.TokenHash == tokenHash);
```
TTA777, line 44: Any reason we are not during the filtering for revoked, and expired in the db?

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
```diff
@@ -220,9 +224,47 @@ private async Task<UserResponse> UserWithRanking(User user)
...
+        /// </summary>
+        /// <param name="tokenHash">The token hash from the magic link</param>
+        /// <returns>A JSON Web Token used to authenticate for other endpoints and a refresh token to re-authenticate without a new magic link</returns>
+        [HttpPost]
+        [AllowAnonymous]
+        [Route("auth")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(MessageResponseDto), StatusCodes.Status404NotFound)]
+        [ProducesDefaultResponseType]
+        public async Task<ActionResult<UserLoginResponse>> Authenticate(string tokenHash)
```
TTA777, line 261: In case we would like to add properties to the request, I'd prefer the tokenhash be wrapped in a request object

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
```diff
@@ -0,0 +1,66 @@
...
+        if (foundToken == null || foundToken.Revoked || foundToken.Expired())
+        {
+            await InvalidateRefreshTokensForUser(foundToken?.User);
+            throw new ApiException("Invalid token", 401);
```
TTA777, line 48: Don't we have a more appropriate exception defined? I haven't checked, but I'd expect we had one

TTA777 (changes requested):
Very minor stuff this time around. Really like the code coverage :)

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
A-Guldborg: Renamed to `GenerateMagicLinkToken` (token of type magic link). https://github.com/AnalogIO/analog-core/pull/289/commits/b051b7c608b57e03e8c8684392b2dbbdcda3a950

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
A-Guldborg: Good question.
I've changed it to be part of the LINQ statement by using a DB function for the Expired field. Note that I keep an implementation (though rewritten to be static as required by EFCore DbFunctions) for client-side evaluation which is used by tests and can be used in the future if we need to call it client-side even though this requires more maintenance. (client being the backend and not database server). Please also help me evaluate if we should place DbFunctions elsewhere?

There is one downside as you can see from the related commit: For the malicious cases where a token is used multiple times it doesn't return a foundToken with a user where we can invalidate that user's tokens. This is only possible if we do it client-side but I am not sure how much we weigh that use-case? If we want to invalidate in this use-case, it only makes sense to do the tokenhash (and maybe Expired) check server-side and the rest client-side. This also leads to a removed test case in https://github.com/AnalogIO/analog-core/pull/289/commits/3d088b85b6a236ed215e3d8211c9791242c2175e (among other fixes found)

Btw, see the following screenshot for the EF Core translation for this implementation:
image

Also, this rabbit-hole led me to fix a bug (in the same commit) ensuring that we hash the generated magic link token before saving it, since we hash it before checking it in the database.

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
A-Guldborg: Couldn't find one more appropriate. This is also how we handle failed authentication attempts in v1 `AccountService`.

coffeecard/CoffeeCard.WebApi/Controllers/v2/AccountController.cs
A-Guldborg: Done in https://github.com/AnalogIO/analog-core/pull/289/commits/d5c5e1203d6fa43717b2509cff10cbeff747be13

A-Guldborg:
@duckth https://github.com/AnalogIO/analog-core/actions/runs/12530002648/job/34946131212?pr=289 fails with some bicep error message that I cannot decode. It seems to indicate (to me) that `module bindCertificate` should *not* depend on `certificate` but I don't know enough about bicep to be sure.

lucasfth (approved):
I would say this is perfectly written code and should be forced merged 🚀

duckth:
> @duckth https://github.com/AnalogIO/analog-core/actions/runs/12530002648/job/34946131212?pr=289 fails with some bicep error message that I cannot decode. It seems to indicate (to me) that `module bindCertificate` should _not_ depend on `certificate` but I don't know enough about bicep to be sure.

Fixed

coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
A-Guldborg: see https://github.com/AnalogIO/analog-core/pull/289/commits/360620b2482ab45e4e7a23d64ce6ff4c2651a2bb

TTA777 (approved):
lgtm

PR #293, opened 2024-11-05
Add EmailSender abstraction for sending mails using either SMTP or Mailgun

duckth:
This PR adds an `IEmailSender` abstraction along with implementations `SmtpEmailSender` and `MailgunEmailSender`, and refactors `EmailService` to use said abstraction. The main idea is to allow for local email testing using an SMTP server such as [MailHog](https://github.com/mailhog/MailHog), which has also been added to the Docker Compose configuration.

image

### DIFF
--- a/coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
+++ b/coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
@@ -1,5 +1,4 @@
 using System.ComponentModel.DataAnnotations;
-using NetEscapades.Configuration.Validation;
 
 namespace CoffeeCard.Common.Configuration
 {
--- a/coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
+++ b/coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
@@ -3,15 +3,10 @@ using NetEscapades.Configuration.Validation;
 
 namespace CoffeeCard.Common.Configuration
 {
-    public class SmtpSettings : IValidatable
+    public class SmtpSettings
     {
         [Required] public string Host { get; set; }
         [Required] public int Port { get; set; }
-
-        public void Validate()
-        {
-            Validator.ValidateObject(this, new ValidationContext(this), true);
-        }
     }
 }
 
--- a/coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
+++ b/coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
@@ -1,0 +1,17 @@
+using System.ComponentModel.DataAnnotations;
+using NetEscapades.Configuration.Validation;
+
+namespace CoffeeCard.Common.Configuration
+{
+    public class SmtpSettings : IValidatable
+    {
+        [Required] public string Host { get; set; }
+        [Required] public int Port { get; set; }
+
+        public void Validate()
+        {
+            Validator.ValidateObject(this, new ValidationContext(this), true);
+        }
+    }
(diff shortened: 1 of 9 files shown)


coffeecard/CoffeeCard.Library/Services/SmtpEmailSender.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace CoffeeCard.Library.Services;
+
+public class SmtpEmailSender(SmtpSettings smtpSettings) : IEmailSender
+{
+    public async Task SendEmailAsync(MimeMessage mail)
+    {
+        mail.From.Add(new MailboxAddress("Cafe Analog", "smtp@analogio.dk"));
+
+        using var smtpClient = new SmtpClient();
+        await smtpClient.ConnectAsync(smtpSettings.Host, smtpSettings.Port);
```
duckth, line 15: This will throw a SocketException if the client cannot connect, e.g. if you are not running a local SMTP client. How should we handle this? Is it a valid assumption that a developer should have a client running, or should we handle the error nicely / ignore it?

coffeecard/CoffeeCard.WebApi/Startup.cs
```diff
@@ -76,6 +76,14 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<Library.Services.v2.IAccountService, Library.Services.v2.AccountService>();
             services.AddScoped<IPurchaseService, PurchaseService>();
             services.AddScoped<IMapperService, MapperService>();
+            if (_environment.IsDevelopment())
```
duckth, line 79: I am realizing that this should be LocalDevelopment as we only want the SMTP client on local, not the remote dev environment

TTA777: Hmm, since we don't override the ASPNETCORE_ENVIRONMENT variable in our deployment, I would expect it to be set to production per default since it is a published application. If you want to use one that we do set, use the one from the EnvironmentSettings.

coffeecard/CoffeeCard.Library/Services/SmtpEmailSender.cs
```diff
@@ -0,0 +1,28 @@
...
+        try
+        {
+            using var smtpClient = new SmtpClient();
+            await smtpClient.ConnectAsync(smtpSettings.Host, smtpSettings.Port);
+            await smtpClient.SendAsync(mail);
+            await smtpClient.DisconnectAsync(true);
+        }
+        catch (Exception ex)
+        {
+            Log.Error($"Error sending mail to SMTP server: {ex.Message}");
```
TTA777, line 25: Use structured logging instead of string interpolation

TTA777: Refer to log statement from the mailgunSender if in doubt

coffeecard/CoffeeCard.Common/Configuration/SmtpSettings.cs
```diff
@@ -0,0 +1,17 @@
+using System.ComponentModel.DataAnnotations;
+using NetEscapades.Configuration.Validation;
+
+namespace CoffeeCard.Common.Configuration
+{
+    public class SmtpSettings : IValidatable
```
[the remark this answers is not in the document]
duckth: We should probably make it optional, as I am also realizing I would need to update the Bicep for deploying the appsettings. But that seems overkill since it is not used. Do you know how I could make it optional or only required in local development? I guess I could change the validator, or just not have it be validated ?

PR #296, opened 2024-11-12
Format DateTime strings as correct ISO8601

### LINKED ISSUE #295: Swipe receipts are displayed in the wrong time zone
**Describe the bug**
Receipts show as swiped an hour earlier than they actually were when viewed in the receipts page.

**To Reproduce**
1. Swipe a ticket and notice the current time
2. Go to the receipts tab
3. View the swipe time on the latest receipt 

**Expected behavior**
Swipe time should be equal to current time

**Device (please complete the following information):**
 - Device: iPhone
 - OS: not available
 - App Version: 3.9.1

**Additional context**

duckth:
Fixes #295 

Turns out that the default string formatting of DateTime objects doesn't conform to ISO8601, so even when the object is saved as UTC; the app does not recognize it as UTC.


coffeecard/CoffeeCard.Tests.Unit/Utils/DateTimeConverterTest.cs
```diff
@@ -0,0 +1,60 @@
...
+            var originalDateTime = new DateTime(2022, 1, 9, 21, 3, 52, DateTimeKind.Utc);
+
+            // Act
+            var json = JsonSerializer.Serialize(originalDateTime, _options);
+            var deserializedDateTime = JsonSerializer.Deserialize<DateTime>(json, _options);
+
+            // Assert
+            Assert.Equal(originalDateTime, deserializedDateTime);
+        }
+    }
```
frederikspang, line 59: Maybe add a test for NON `DateTimeKind.Utc` to check handling.

frederikspang (approved):

TTA777 (approved):

Issue #298, opened 2024-11-12
Fix SonarCloud coverage


duckth:
Currently SonarCloud is mistakenly reporting 0% coverage on our project. Fix the scanner to correctly report coverage percentage

PR #299, opened 2024-11-12
Update SonarCloud workflow, fix .net8 path

duckth:
Updates the SonarCloud workflow to some newer versions (mostly taken from the template given by SonarCloud themselves), and hopefully fixes the coverage issue in #298 by fixing the path by changing .net6 -> .net8


TTA777 (approved):
Let's give it a go

PR #300, opened 2024-11-17
Remove MobilePay RefundConsoleApp project

marfavi:


marfavi:
Need to remove more references first

marfavi:
Already removed

PR #302, opened 2024-11-17
Feature/otel

TTA777:


infrastructure/prd.settings.json
duckth, line 1: What about dev settings? Do we still use regular Azure logging for dev?

TTA777: It was left out on purpose, yes, since we only have a single deployment for grafana for now and we wouldn't have a way of identifying if it was prd or dev that sent the log. Might be nicer to include the environment as part of the data though, then we should be able to do it with just the one deployment

duckth: Yeah I'd like if we could enrich the log entry with the environment too. I assume we would then be able to create filters in grafana to search each environment. But it might be a bit overkill given the amount of times I have had to dig in the dev logs.

marfavi:
What does this PR do?

infrastructure/prd.settings.json
TTA777: They are now enriched with a resource attribute that should apply to all metrics, traces, and logs

duckth (approved):

PR #304, opened 2024-11-26
Changing v2 account service tests to use builders

A-Guldborg:
This pull requests addresses @TTA777 review on #289 wrt using test data builders.

This is still a WIP/draft since tests are failing. Please see https://github.com/AnalogIO/analog-core/pull/289#issuecomment-2502152638 for a thorough explanation of the issue.

This PR could also be responsible for adding integration test(s) & addressing the v2 `TokenServiceTests` added in #289

### DIFF
--- a/coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
+++ b/coffeecard/CoffeeCard.Library/Services/v2/TokenService.cs
@@ -22,7 +22,7 @@ public class TokenService : ITokenService
     public async Task<string> GenerateMagicLink(User user)
     {
         var guid = Guid.NewGuid().ToString();
-        var magicLinkToken = new Token(guid, TokenType.MagicLink);
+        var magicLinkToken = new Token { TokenHash = guid, Type = TokenType.MagicLink };
 
         user.Tokens.Add(magicLinkToken);
         await _context.SaveChangesAsync();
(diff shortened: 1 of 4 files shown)


TTA777:
Hmm, it seems to me that the simplest solution to your conundrum @A-Guldbor,g is to go for option two, and making the builder use the constructor you want. 

I tried playing a bit locally, and the following works for me 
``` csharp
public static TokenBuilder Simple()
{
var builder = new TokenBuilder();
            builder.Faker.CustomInstantiator(f => new Token("some token hash here, maybe use the fake to generate it", TokenType.Refresh));
            return builder
                .WithExpires(DateTime.Now.AddDays(1))
                .WithRevoked(false)
                .WithUser(UserBuilder.DefaultCustomer().Build());
}
```

I'm unsure at present what you dislike about this solution, since it seems we can keep using the builders the way we currently are, and you can keep the current code design. 

My only alternative, off the top of my head, would be to:
- Make the setter for Expires private and adding logic to the TokenType setter ensuring it sets the right expiration based on the type as well
- Mark both the tokenHash and type as required as you started to do in this PR

This approach would probably have one problem though, being you wouldn't be able to manipulate the exipration of a token for a test. I'm unsure if that is something we would even want to do, but it sounds reasonably within the scope of things we might do in a test.

As to how the builders actually work, well the short answer is probably a bit of black magic. More seriously, it is making use of a [source generator](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md) (our builderGenerator.cs), and a library called [Bogus](https://github.com/bchavez/Bogus). The generator makes sure we have the WithPropteryName methods, and the bogus library can then be used for creating somewhat realistic test data when we want to, instead of being specific.

PR #357, opened 2026-05-01
Migrate to swashbuckle

marfavi:
might write a longer description soon

allows for proper polymorphism in generated swagger definitions


marfavi:
#359 does it better

PR #366, opened 2026-08-27
feat: set profile picture

marfavi:


coffeecard/CoffeeCard.Library/Services/AccountService.cs
```diff
@@ -177,6 +177,11 @@ public async Task<User> RegisterAccountAsync(
             _context.Users.Add(user);
             await _context.SaveChangesAsync();
 
+            // Assign profile icon and background color using the same rules as the UserIcon widget in coffeecard_app
+            user.ProfileIcon = (ProfileIcon)(user.Id % 9);
```
TTA777: Consider making these two dynamic so it calculates the number based on the length of the enum. That way new use will automatically start using any new icons

coffeecard/CoffeeCard.Library/Services/v2/LeaderboardService.cs
```diff
@@ -38,6 +38,10 @@ int top
                         Name = s.User.PrivacyActivated ? "Anonymous" : s.User.Name,
                         Rank = index + 1,
                         Score = s.SwipeCount,
+                        ProfileIcon = s.User.PrivacyActivated ? null : s.User.ProfileIcon,
```
TTA777: Not checks not needed?

coffeecard/CoffeeCard.Library/Migrations/20260413154037_AddProfilePicToUser.cs
```diff
@@ -0,0 +1,63 @@
...
+                nullable: true);
+
+            // Assign values to existing users using the same rules as the UserIcon widget in coffeecard_app:
+            // icon = id % 9, background color = id % 10
+            migrationBuilder.Sql(
+                "UPDATE [dbo].[Users] SET [ProfileIcon] = [Id] % 9");
+            migrationBuilder.Sql(
+                "UPDATE [dbo].[Users] SET [ProfileBackgroundColor] = [Id] % 10");
+
+            migrationBuilder.AlterColumn<int>(
```
TTA777: I would prefer if we could get ef to generate these two by itself in a sepearete migration to ensure it does not get confused about the state of the db

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -92,6 +92,11 @@ int programme
             _context.Users.Add(user);
             await _context.SaveChangesAsync();
```
TTA777: You probably want to remove this line, as you added another saveChanges further down

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -174,6 +179,34 @@ public async Task<User> UpdateAccountAsync(User user, UpdateUserRequest updateUs
...
+                user.ProfileIcon = updateUserRequest.ProfileIcon.Value;
+                _logger.LogInformation(
+                    "User changed profile icon to {profileIcon}",
+                    user.ProfileIcon
+                );
+            }
+
+            if (updateUserRequest.ProfileBackgroundColor != null)
+            {
+                if (!Enum.IsDefined(updateUserRequest.ProfileBackgroundColor.Value))
```
TTA777: Will the api not always enforce this?

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
```diff
@@ -174,6 +179,34 @@ public async Task<User> UpdateAccountAsync(User user, UpdateUserRequest updateUs
                 _logger.LogInformation("User changed password");
             }
 
+            if (updateUserRequest.ProfileIcon != null)
+            {
+                if (!Enum.IsDefined(updateUserRequest.ProfileIcon.Value))
```
TTA777: Will the api not always enforce this?

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Leaderboard/LeaderboardEntry.cs
```diff
@@ -36,6 +39,20 @@ public sealed class LeaderboardEntry : IEquatable<LeaderboardEntry>
         /// <example>2</example>
         public required int Rank { get; set; }
 
+        /// <summary>
+        /// User's profile icon
+        /// </summary>
+        /// <value>Profile Icon</value>
+        /// <example>Teabag</example>
+        public required ProfileIcon? ProfileIcon { get; set; }
```
TTA777: not nullable

coffeecard/CoffeeCard.Models/DataTransferObjects/v2/Leaderboard/LeaderboardEntry.cs
```diff
@@ -36,6 +39,20 @@ public sealed class LeaderboardEntry : IEquatable<LeaderboardEntry>
...
+
+        /// <summary>
+        /// User's profile picture background color
+        /// </summary>
+        /// <value>Background Color</value>
+        /// <example>MossGreen</example>
+        public required ProfileBackgroundColor? ProfileBackgroundColor { get; set; }
```
TTA777: not nullable

TTA777 (changes requested):

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
marfavi: we save to generate the user id, which is used below

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
marfavi: no. Startup.cs registers JsonStringEnumConverter with the default allowIntegerValues: true, so out-of-range integers deserialize fine and reach the service.

coffeecard/CoffeeCard.Library/Services/v2/AccountService.cs
marfavi: no. Startup.cs registers JsonStringEnumConverter with the default allowIntegerValues: true, so out-of-range integers deserialize fine and reach the service.

